PR #4, opened 2021-09-23
Specify distance ranges for BEIS flights

greenercode:
**What?**

Specify distance ranges for BEIS flights

**Why?**

Prevent unclarity between to/from UK, international and domestic flights


patstirling (commented):
- suggest IDs be changed to specify inclusion of contrails, with a consistent format (e.g. short-haul-flight-with-contrail OR short-haul-flight-without-contrail)
- need to change description to reflect the above. Please review other UK Air Travel lines (1154, 1155, 1156, 1157, 1158, 1159) to ensure consistency

PR #5, opened 2021-09-23
Provide better identification for flights with/without contrails

greenercode:
**What?**

Clarify emissions for air travel, provide clear IDs.

**Why?**

Making the distinction between EFs that include indirect effects of non-CO2 emissions and those that include direct effects of CO2, CH4 and N2O emissions only (change within the description).


patstirling (changes requested):
There is a larger problem here of consistency of IDs for air travel. Suggest we change them in this pull request. To wit:

Suggest all IDs in the air travel category be changed to a standardised format which clarifies:

- [haul length]
- [domestic/international]
- [travel class/average]
- [with-contrails/without-contrails]

Please review all Air Travel lines to ensure consistency in both IDs and descriptions

grindcode:
The currently used assumption is that contrails should be taking into account as a default.

For this reason, an option to go about this:
- Provide no further specification for EFs that contain contrails in the methodology.
- Provide `-without-contrails` as a suffix for EFs that do NOT contain contrails in the methodology.

Open for discussion.

patstirling:
> The currently used assumption is that contrails should be taken into account as a default.
> 
> For this reason, an option to go about this:
> 
> * Provide no further specification for EFs that contain contrails in the methodology.
> * Provide `-without-contrails` as a suffix for EFs that do NOT contain contrails in the methodology.
> 
> Open for discussion.

Agreed - this is neater and defaults to the reality (forces people to deliberately choose the less complete version

patstirling (changes requested):
Line 418 - does this need "uk-" at beginning? Region is GB so this specifies country. I think "domestic-flight-average-passenger

Line 967 - think "to-from-non-uk-" is a bit laboured; "international-outside-uk" would be easier to understand. Note 966, 968, 969 and 975 are also "international-outside-uk" according to their descriptions

Line 1000, 1003-1006, 1152-1154, 1158-1160 - can we conform to format here e.g. "long-haul-flight-international-average-passenger-without-contrails". See note above about "uk-" at beginning. Note position of "flight" should be consistent

Can we also update air travel EFs from other sources to align with the format above. Eg. 10, 419-422, 970-974, 976-978, 1001-1002, 1032-1033, 1155-1157 & 1303-1304

maryarbab (approved):

OpenEmissionFactorsDB.csv
```diff
@@ -1,1303 +1,1304 @@
...
-Energy,Electricity,grid-mix,Grid Mix,kg-CO2e/kwh,0.264,10,ADEME,2014,VE,8/13/21,Emission Intensity of Grid Mix. Venezuela. Updated 2014. Published by the Agence de l'Environnement et de la Mątrise de l'Énergie. Manufactured in France. Provided by the IEA.,https://www.bilans-ges.ademe.fr/en/accueil
-Energy,Electricity,grid-mix,Grid Mix,kg-CO2e/kwh,0.298,10,ADEME,2014,PA,8/13/21,Emission Intensity of Grid Mix. Panama. Updated 2014. Published by the Agence de l'Environnement et de la Mątrise de l'Énergie. Manufactured in France. Provided by the IEA.,https://www.bilans-ges.ademe.fr/en/accueil
-Energy,Electricity,grid-mix,Grid Mix,kg-CO2e/kwh,0.72,10,ADEME,2014,BJ,8/13/21,Emission Intensity of Grid Mix. Benin. Updated 2014. Published by the Agence de l'Environnement et de la Mątrise de l'Énergie. Manufactured in France. Provided by the IEA.,https://www.bilans-ges.ademe.fr/en/accueil
-Energy,Electricity,grid-mix,Grid Mix,kg-CO2e/kwh,0.22,10,ADEME
```
patstirling, line 963: this should read "non-uk countries" and we should make clear that BEIS does not make any distinctions between short-, medium- or long-haul flights; we should add this in the descriptions of all lines that this applies to

patstirling (changes requested):

patstirling (approved):

PR #12, opened 2021-09-30
Specify ADEME electricity emission factors

greenercode:
**What?**

Specify ADEME electricity emission factors.

**Why?**

Lack of clarity on combustion and/or upstream in emission factor description.


patstirling (changes requested):
See suggested changes in review. General questions @greenercode @maryarbab - is "upstream" the right term here... or "indirect"... or something else. We have a discovery piece which is aiming to define this, should we be waiting until that is complete?

PR #21, opened 2021-10-17
Update structure for Accommodation 

maryarbab:

- [x] Update

Electricity

**What?**

Updates all current electricity IDs to adhere to the following structure:

non-pv EFs:

electricity-energy_source_*[e.g. "grid_mix" or "coal_fired_plant"]*-activtity_*[e.g. "fuel_combustion", "upstream" or "fuel_combustion_and_upstream"]*

pv EFs:
electricity-energy_source_[e.g. "grid_mix" or "coal_fired_plant"]-activtity_[e.g. "fuel_combustion", "upstream" or "fuel_combustion_and_upstream"]-region_produced_[e.g. "China"]

In addition to IDs, names, descriptions, and factors are also updated.

**Why?**

This is part of a larger refactoring of IDs across the OEFDB to adhere to a specified structure per category, allowing them to be better understood both by humans and machines.


patstirling (changes requested):
The data itself all looks spot-on, however the ID and category fields appear to have switched. There is also a question about the categorisation of hotel stays which needs to be assessed separately (more detail to come on resolution of this).

maryarbab:
Great observation, Pat! Just made a new commit to address the issue. Looking forward to our discussion regarding the travel category. @patstirling

PR #22, opened 2021-10-18
IDs changed for the Storage (Computing) Category

AnastasiaLobanova:

- [ ] New
- [x] Update

Storage(Computing)

**What?**

Updates all current Storage (Computing) IDs to adhere to the following structure:

storage-provider_[e.g. *"aws"*]-region_[e.g. *"af_south_1"*]-type_[e.g. *"hdd"*]

an example:

storage-provider_aws-region_af_south_1-type_hdd

**Why?**

This is part of a larger refactoring of IDs across the OEFDB to adhere to a specified structure per category, allowing them to be better understood both by humans and machines.


maryarbab (changes requested):
Great update! The category looks very clean and polished:) Just one tiny note: can we change "Using the assumptions and grid mix data available in the source" to "Using the assumptions and grid emissions factors available in the source" to be accurate?

PR #38, opened 2021-11-02
IDs are updated for Rail Transport category

AnastasiaLobanova:

- [ ] New
- [x] Update

Rail Transport

**What?**

IDs for the rail transport category changed as:

freight_train-route_type_[e.g. domestic or unknown]-fuel_type_[e.g. diesel]-activity_[e.g. fuel_combustion]

**Why?**

This is part of a larger refactoring of IDs across the OEFDB to ensure EFs adhere to a clear standard structure per category and across sources, allowing them to be better understood and compared both by humans and machines.


greenercode (approved):

OpenEmissionFactorsDB.csv
```diff
@@ -1242,8 +1242,8 @@ Transport,Sea Transport,sea_freight-vessel_type_overseas_goods_ship-route_type_u
...
-Transport,Rail Transport,rail-freight-transport-diesel,Rail freight transport diesel,kg-CO2e/tkm,0.034,,UBA,2020,DE,2021/09/15,Emission intensity of rail freight transport. Diesel fuel. Data from the Öko-Institut's renewbility technology database. Value applicable to 2020.Provided by the Öko-Institut. Accessed on the 15th of September 2021.,https://www.probas.umweltbundesamt.de/php/prozessdetails.php?id={9B2D7023-0BCA-4F73-BC74-3A6B98DE7D48}
+Transport,Rail Transport,freight_train-route_type_domestic-fuel_type_electric-activity_upstream_and_fuel_combustion,Rail freight electric train ,kg-CO2e/tkm,0.019,,UBA,2020,DE,2021/09/15,Emission intensity of goods transported by electric rail freight transport. Fuel combustion and upstream LCA activity is included. Data from the Öko-Institut's renewbility technology database. Value applicable to 2020.Provided by the Öko-Institut. Accessed on the 15th of September 2021.,https://www.probas.umweltbundesamt.de/php/prozessdetails.php?id={CEB3265F-2B8B-46EA-9166-4BD4F07EAD1F}
```
maryarbab, line 1245: Does the source specify what type of upstream emission that is? Fuel upstream?

maryarbab: We need to check that we have used consistent wording for the LCA terms in the activity column as well as the description. This can be done as we have all PRs merged.

OpenEmissionFactorsDB.csv
```diff
@@ -455,7 +455,7 @@ Transport,Sea Travel,passenger_ferry-route_type_foot_passenger-fuel_source_unkno
...
-Transport,Rail Transport,freight-train,Freight Train,kg-CO2e/tkm,0.02782,,BEIS,2021,GB,2021/07/23,Emission Intensity of Freight Train. Available in DEFRA's Greenhouse Gas Reporting: Conversion Factors 2021.,https://www.gov.uk/government/publications/greenhouse-gas-reporting-conversion-factors-2021
+Transport,Rail Transport,freight_train-route_type_unknown-fuel_type_unknown-activity_fuel_combustion,Rail freight train,kg-CO2e/tkm,0.02782,,BEIS,2021,GB,2021/07/23,Emission intensity of goods transported by rail freight transport. Based on a mixture of emissions from diesel and electric rail. Only fuel combustion activity is included. Available in BEIS Greenhouse Gas Reporting: Conversion Factors 2021.,https://www.gov.uk/government/publications/greenhouse-gas-reporting-conversion-factors-2021
```
maryarbab, line 458: Does "emissions from electric rail" make sense? how about "a mixture of emissions from diesel and electricity generation"?

maryarbab (changes requested):

OpenEmissionFactorsDB.csv
greenercode: Agreed.

OpenEmissionFactorsDB.csv
AnastasiaLobanova: I will change it!

OpenEmissionFactorsDB.csv
AnastasiaLobanova: According to UBA included are all the upstream activities (but they do not specify the system boundaries explicitly)

greenercode: https://www.probas.umweltbundesamt.de/php/prozessdetails.php?id={CEB3265F-2B8B-46EA-9166-4BD4F07EAD1F} The given EF does not provide further information, except that it corresponds to more than simply fuel combustion;  _"The direct environmental effects plus those emanating from the upstream process chains. In each case, the system boundaries must be observed, which may be set differently depending on the data source (e.g., consideration of production and disposal, credit_s)." More information required for both data source and specific emission factor calculation at a later stage.

maryarbab: Cool, thank you!

maryarbab (dismissed):

OpenEmissionFactorsDB.csv
patstirling: This is still confusing: "Based on a mixture of emissions from diesel and electricity generation. Only fuel combustion activity is included." this implies that despite electric trains being included in the mix, only the combustion of diesel fuel is taken into account. Better would be "Only emissions from fuel combustion or generation of electricity consumed are included"

OpenEmissionFactorsDB.csv
patstirling: This line also should not refer to "combustion" as it is electric - rather: "Direct electricity consumption and upstream LCA activity is included"

patstirling (changes requested):
See comments in-line

greenercode (dismissed):

greenercode (approved):

patstirling (approved):

PR #52, opened 2021-11-10
Resolve esoteric IDs in the Rail Travel

maryarbab:

- [ ] New
- [x] Update

Rail Travel

**What?**

Resolve esoteric IDs in the fuel category

**Why?**

Provide better comparability across EFs


AnastasiaLobanova (approved):

patstirling (changes requested):
Quite a lot of (small) changes outside the rail travel category - these are likely to cause conflicts

patstirling (approved):

PR #79, opened 2021-12-01
Adding LCA activity column

maryarbab:

-  New
- [ ] Update

All

**What?**

Adding LCA activity for each emission factor based on the information that the source provides.

**Why?**

The user knows exactly what each emission factor refers to in terms of LCA activity.


OpenEmissionFactorsDB.csv
```diff
@@ -362,68 +362,68 @@ Transport,Sea Transport,sea_freight-vessel_type_container-route_type_na-vessel_l
...
-Transport,Road Travel,passenger_vehicle-vehicle_type_bus-fuel_source_bev-distance_na-engine_size_na,Wellington electric bus,kg-CO2e/pkm,0.01250848,,MfE,2020,NZ,2021/09/10,Emission intensity of Wellington electric bus. Provided by the Ministry for the Environment. Published by the Ministry for the Environment of New Zealand in 2020.,https://environment.govt.nz/publications/measuring-emissions-detailed-guide-2020/,fuel_combustion
+Transport,Road Travel,passenger_vehicle-vehicle_type_city_bus-fuel_source_bev-distance_na-engine_size_na,Electric city bus,kg-CO2e/pkm,0.0217,70,ADEME,2018,FR,2021/08/20,Emission Intensity of electric city bus. Continental France. 2020 Factor. Updated April 2020. Published by ADEME. Provided by GT Transport Base Carbone & HBEFA.,https://www.bilans-ges.ademe.fr/en/accueil,upstream-vehicle_manufactuing
```
patstirling, line 425: manufactuing -> manufacturing

OpenEmissionFactorsDB.csv
```diff
@@ -19,37 +19,37 @@ Transport,Road Travel,passenger_vehicle-vehicle_type_car-fuel_source_na-distance
...
-Energy,Fuel,fuel_type_avtur-fuel_use_aviation,Aviation turbine fuel (avtur),kg-CO2e/l,2.54514,,BEIS,2021,GB,2021/07/23,Emission intensity of aviation turbine fuel for turbo-prop aircraft and jets (aka jet fuel). Similar to kerosene used as a heating fuel but refined to a higher quality. The fuels conversion factors should be used for primary fuel sources combusted at a site or in an asset owned or controlled by the reporting organisation. Available in BEIS Greenhouse Gas Reporting: Conversion Factors 2021.,https://www.gov.uk/government/publications/greenhouse-gas-reporting-conversion-factors-2021,combustion
+Transport,Vehicle,passenger_vehicle-vehicle_type_car-fuel_source_fcev_smr_centralized_natural_gas-engine_size_na-vehicle_age_na-vehicle_weight_na,Hydrogen car (SMR centralized natural gas),kg-CO2e/km,0.216,70,ADEME,2018,FR,2021/08/20,Emission intensity of car fuelled by hydrogen produced by SMR (Small Modular Reactor) centralized natural gas. Updated April 2020. Published by ADEME. Provided by the HBEFA & Yearly Compte des Transports.,https://www.bilans-ges.ademe.fr/en/accueil,fuel_upstream-vehicle_manufacturing
```
patstirling, line 47: Is "vehicle_manufacturing" too specific? Could we use "manufacturing" as a more general term, as I imagine we'll have EFs reflecting the manufacture of all sorts of things in the future.

patstirling (changes requested):
Completely amazing work! Couple of thoughts in comments - also, could you update the DATA_GUIDANCE doc within the pull request to reflect these changes?

AnastasiaLobanova (changes requested):
Thank you Maryam!! Great piece of work! Should be very helpful for the user to orient and also will save loooots of time for the users.

I have two general comments:

I suggest no spaces in header - better for the automatic processing i.e. LCA activity -> LCA_activity

I am not 100% clear on how the hyphens and underscores are used? Hyphens to separate different LCA phases? Then I don't understand this one: upstream-vehicle_manufacturing - here the vehicle manufacturing is not upstream? Or what is meant by upstream here? 


Small comment: 

There are some extra spaces at the end of the "upstream-vehicle_manufacturing "

OpenEmissionFactorsDB.csv
maryarbab: Good point. Actually, ADEME has both "vehicle manufacturing" and "manufacturing" for EFs. I decided to have it all as "vehicle manufacturing" to be consistent as I could not think of any other manufacturing activity for these factors. But I can change that to "manufacture" to be inclusive. The other point is that in this column we are trying to keep activities as close as possible to the source and then group them in the LCA boundary column. Any thoughts?

AnastasiaLobanova (dismissed):

patstirling (changes requested):
small change to column header - also to "manufacturing" as discussed

patstirling (dismissed):

patstirling (approved):

PR #81, opened 2021-12-03
Re-order DB alphabetically by sector > category > name

patstirling:

- [ ] New
- [X] Update

All

**What?**

Re-orders the database alphabetically by `sector`, then `category`, then `name`

**Why?**

The database is currently unordered, which makes it difficult to manage when reviewing/editing


grindcode (approved):

maryarbab (approved):
Looks great!

PR #85, opened 2021-12-06
Include fugitive gas conversion factors

greenercode:

- [x] New

Fugitive gas emissions

**What?**

Support fugitive gas emissions (conversion factors) from BEIS. 

**Why?**

Support fugitive gas emissions calculation. These conversion factors should enable users to report; 1. leakage from air-conditioning, refrigeration units 2. The release to the atmosphere of other gases that have global warming potential (GWP).


patstirling (changes requested):
- sector and category should be "Refrigerants and Fugitive Gases" as discussed
- ID format suggest: "fugitive_gas-type_halon_1211-use_na"
- should we include "use_"? Do we need this subID? Is it anticipated that there will be EFs in this category that we will need to specify the use of? This may be sensible for future-proofing if we anticipate that EFs may be different depending on the specific use of the gas - but I don't think that's likely given this is a value that relates directly to the gas itself
- remove commas from the description
- source link should ideally be a webpage (eg. https://www.gov.uk/government/publications/greenhouse-gas-reporting-conversion-factors-2021) rather than a file, with indication in description as to which file the EF was retrieved from (e.g. "Available in BEIS Greenhouse Gas Reporting: Conversion Factors 2021." -> "Retrieved from BEIS Greenhouse Gas Reporting Conversion Factors 2021: flat file." (if they were retrieved from the flat file)

patstirling (changes requested):
Thanks for the updates @greenercode - could you please address the below:

- ID field - replace "fugitive_gas_type_" with "fugitive_gas-type_"
- Description to read: "Retrieved from BEIS Greenhouse Gas Reporting Conversion Factors 2021: flat file." <- specifically refers to the file used
- Another update to the description field; as we are very close to the BEIS guidance anyway, we may as well use this verbatim from their guidance: "...used to report leakage from air-conditioning, refrigeration units or the release to the atmosphere of other gases that have global warming potential (GWP)."
- please reorder the database by name then category then sector

OpenEmissionFactorsDB.csv
```diff
@@ -791,6 +791,93 @@ Organisational activities,Government Activities,organisational_activity-type_gro
...
+Refrigerants and Fugitive Gases,Refrigerants and Fugitive Gases,fugitive_gas-type_hfc_41 ,HFC-41,kg-CO2e/kg,92,,BEIS,2021,GB,2021/11/26,"Emission intensity for HFC-41. This conversion factor should be used to report leakage from air-conditioning, refrigeration units or the release to the atmosphere of other gases that have global warming potential (GWP).Retrieved from BEIS Greenhouse Gas Reporting Conversion Factors 2021: flat file.",https://www.gov.uk/government/publications/greenhouse-gas-reporting-conversion-factors-2021,unknown
+Refrigerants and Fugitive Gases,Refrigerants and Fugitive Gases,fugitive_gas-type_hfc_43_I0mee ,HFC-43-I0mee,kg-CO2e/kg,1640,,BEIS,2021,GB,2021/11/26,"Emission intensity for HFC-43-I0mee. This conversion factor should be used to report leakage from air-conditioning, refrigeration units or the release to the atmosphere of other gases that have global warming potential (GWP).Retrieved from BEIS Greenhouse Gas Reporting Conversion Factors 2021: flat file.",https://www.gov.uk/government/publications/greenhouse-gas-reporting-conversion-factors-2021,unknown
```
maryarbab, line 833: Great job, @greenercode! Just a tiny note: is "I0mee" correct?

maryarbab (changes requested):

OpenEmissionFactorsDB.csv
greenercode: @maryarbab Yes that is correct!

maryarbab (dismissed):

patstirling (approved):

PR #90, opened 2021-12-09
Convert EPA data to reflect CO2 equivalencies

greenercode:

- [x] Update

EPA data source emission factors (excluding electricity). 

**What?**

Before, these were only reflecting CO2 Factor (kg / unit). Now these include CH4 Factor (g / unit), N2O Factor (g / unit), which post unit- conversion reflect metric CO2 equivalencies.

**Why?**

At the moment, the OEFDB strives to facilitate UX by providing CO2 equivalencies. In the future, gases might be included individually according to their GWP (global warming potential).


AnastasiaLobanova (commented):
Thank you Stephane! Good job!! And which GWP factors you have used?

maryarbab (commented):
Great job! Thank you for correcting this, @greenercode!

maryarbab (approved):

PR #91, opened 2021-12-09
Sea transport erroneous units

greenercode:

[x] Update

Sea transport 

**What?**

A couple of identified erroneous units, essentially coming from this source https://www.smartfreightcentre.org/pdf/GLEC-report-on-GHG-Emission-Factors-for-Inland-Waterways-Transport-SFC2018.pdf. For one BEIS (UK) and one MfE (Nz) EF units were in g-CO2e/tkm instead of kg-CO2e/tkm. I also reviewed the LCA activity format for Well_to_Propeller to well_to_propeller.

**Why?**

The values were incorrect from a scientific perspective.


AnastasiaLobanova (changes requested):
Is this correct that we have in some cases g-CO2e and in some kg? 

UPD: Ah now I see from the factor value :) Sorry

AnastasiaLobanova (approved):

maryarbab (changes requested):
I am still confused: why don't we make the conversion to only have one unit (i.e. kg of CO2eq)?

greenercode:
> I am still confused: why don't we make the conversion to only have one unit (i.e. kg of CO2eq)?

I can do! But should we not stay truthful to the source considering this is g-CO2e/tkm (just not Kg-CO2e/tkm)? Perhaps this can be easily converted in the API.

OpenEmissionFactorsDB.csv
```diff
@@ -1164,7 +1164,7 @@ Transport,Sea Transport,sea_freight-vessel_type_container-route_type_na-vessel_l
...
-Transport,Sea Transport,sea_freight-vessel_type_container_coupled_convoy-route_type_na-vessel_length_na-tonnage_na-fuel_source_na,Container vessels coupled convoy,kg-CO2e/tkm,19.7,,STC-Nestra B.V.,2020,EU,2021/08/20,Emission intensity for a container vessel coupled convoy. Rhine route. Published by STC-NESTRA B.V in 2018.,https://www.smartfreightcentre.org/pdf/GLEC-report-on-GHG-Emission-Factors-for-Inland-Waterways-Transport-SFC2018.pdf,upstream-fuel_combustion
+Transport,Sea Transport,sea_freight-vessel_type_container_coupled_convoy-route_type_na-vessel_length_na-tonnage_na-fuel_source_na,Container vessels coupled convoy,g-CO2e/tkm,19.7,,STC-Nestra B.V.,2020,EU,2021/08/20,Emission intensity for a container vessel coupled convoy. Rhine route. Published by STC-NESTRA B.V in 2018.,https://www.smartfreightcentre.org/pdf/GLEC-report-on-GHG-Emission-Factors-for-Inland-Waterways-Transport-SFC2018.pdf,upstream-fuel_combustion
```
maryarbab, line 1167: Did we finally decide to have one unit here and convert units or have multiple ones?

maryarbab (changes requested):

OpenEmissionFactorsDB.csv
greenercode: Not sure! @maryarbab @patstirling @AnastasiaLobanova do we have a consensus about this? Going by the normalisation argument of the DB we could in fact do the conversions?

AnastasiaLobanova: I agree with both! Question is - if API is doing it then we do not need to normalise it by hand in my view

patstirling: Let's stay true to the source here (as long as it's correct); the API will take the thinking out of it for folks using it to access the full range of EFs

patstirling (approved):

maryarbab (approved):

PR #92, opened 2021-12-09
Update UBA train travel name and descriptions

greenercode:

- [x] Update

Rail travel

**What?**

Update UBA train travel name and descriptions 

**Why?**

These used to indicate near and remote. This has been modified to local and long distance to improve UX.


AnastasiaLobanova (approved):

OpenEmissionFactorsDB.csv
```diff
@@ -990,8 +990,8 @@ Transport,Rail Transport,freight_train-route_type_domestic-fuel_source_electrici
...
-Transport,Rail Travel,passenger_train-route_type_local-fuel_source_diesel,Diesel near passenger train,kg-CO2e/pkm,0.0713,,UBA,2020,DE,2021/09/15,Emission intensity of diesel local passenger train including upstream. Data from the Öko-Institut.,https://www.probas.umweltbundesamt.de/php/prozessdetails.php?id={08984DE8-B8F1-46C1-824B-910FD9E8A023},upstream-fuel_combustion
-Transport,Rail Travel,passenger_train-route_type_long_distance-fuel_source_diesel,Diesel remote passenger train,kg-CO2e/pkm,0.044,,UBA,2020,DE,2021/09/15,Emission intensity of long-distance diesel passenger train including upstream. Data from the Öko-Institut.,https://www.probas.umweltbundesamt.de/php/prozessdetails.php?id={968DDFF4-52FF-4C77-8F1D-F21DD202C11C},upstream-fuel_combustion
+Transport,Rail Travel,passenger_train-route_type_local-fuel_source_diesel,Local passenger train (diesel),kg-CO2e/pkm,0.0713,,UBA,2020,DE,2021/09/15,Emission intensity of local passenger train (diesel) including upstream. Data from the Öko-Institut.,https://www.probas.umweltbundesamt.de/php/prozessdetails.php?id={08984DE8-B8F1-46C1-824B-910FD9E8A023},upstream-fuel_combustion
```
maryarbab, line 993: I would suggest having it as "Diesel local passenger train" and "Diesel long-distance passenger train" to be consistent with the rest (e.g. row 995).

maryarbab (changes requested):

OpenEmissionFactorsDB.csv
greenercode: @AnastasiaLobanova @patstirling Long- it would be great to have a consensus on order here. I agree with the @maryarbab's consistency argument vis-à-vis the rest of the OEFDB, but not sure whether diesel is the activity? This is why I do not think it should go first in the name column. Please share your thoughts.

patstirling: The standard form from name across the transport sector is to have the fuel type followed by the vehicle type - e.g. "Diesel bus" - suggest we conform to that with "Diesel local passenger train". In description "Emission intensity of diesel local passenger train including fuel combustion and upstream" would be ideal.

greenercode: 👌

OpenEmissionFactorsDB.csv
```diff
@@ -1009,8 +1009,8 @@ Transport,Rail Travel,passenger_train-route_type_na-fuel_source_electricity,Elec
...
+Transport,Rail Travel,passenger_train-route_type_local-fuel_source_electricity,Electric local passenger train,kg-CO2e/pkm,0.0546,,UBA,2020,DE,2021/09/15,Emission intensity of electric local passenger train including upstream and electricity consumption. Data from the Öko-Institut.,https://www.probas.umweltbundesamt.de/php/prozessdetails.php?id={5A001C57-5EA0-4301-8139-575EC2E58232},upstream-electricity_consumption
+Transport,Rail Travel,passenger_train-route_type_long_distance-fuel_source_electricity,Electric long-distance passenger train,kg-CO2e/pkm,0.00948,,UBA,2020,DE,2021/09/15,Emission intensity of electric long-distance passenger train including upstream and electricity consumption. Data from the Öko-Institut.,https://www.probas.umweltbundesamt.de/php/prozessdetails.php?id={A4FFA0CD-2550-4435-BBDE-00C6F1A0B22F},upstream-electricity_consumption
```
maryarbab, line 1013: Does the source mention "electricity consumption"?

maryarbab (changes requested):

OpenEmissionFactorsDB.csv
patstirling: Opening gambit is that this should align with the "electricity_generation" value for all other electric trains/vehicles in the DB. It's a valid question as to whether generation, T&D and extraction/fuel transportation is included in "electricity_generation" for UBA EFs though - for example BEIS don't include T&D (that's a separate EF)

maryarbab: Yes exactly: electricty_consumption happens at the end-user side, taking into account the T&D losses, while electrcity_generation happens at the power plant before any transmission and distribution to the end-user.

OpenEmissionFactorsDB.csv
```diff
@@ -990,8 +990,8 @@ Transport,Rail Transport,freight_train-route_type_domestic-fuel_source_electrici
...
-Transport,Rail Travel,passenger_train-route_type_local-fuel_source_diesel,Diesel near passenger train,kg-CO2e/pkm,0.0713,,UBA,2020,DE,2021/09/15,Emission intensity of diesel local passenger train including upstream. Data from the Öko-Institut.,https://www.probas.umweltbundesamt.de/php/prozessdetails.php?id={08984DE8-B8F1-46C1-824B-910FD9E8A023},upstream-fuel_combustion
-Transport,Rail Travel,passenger_train-route_type_long_distance-fuel_source_diesel,Diesel remote passenger train,kg-CO2e/pkm,0.044,,UBA,2020,DE,2021/09/15,Emission intensity of long-distance diesel passenger train including upstream. Data from the Öko-Institut.,https://www.probas.umweltbundesamt.de/php/prozessdetails.php?id={968DDFF4-52FF-4C77-8F1D-F21DD202C11C},upstream-fuel_combustion
+Transport,Rail Travel,passenger_train-route_type_local-fuel_source_diesel,Diesel local passenger train,kg-CO2e/pkm,0.0713,,UBA,2020,DE,2021/09/15,Emission intensity of diesel local passenger train including upstream and fuel combustion. Data from the Öko-Institut.,https://www.probas.umweltbundesamt.de/php/prozessdetails.php?id={08984DE8-B8F1-46C1-824B-910FD9E8A023},electricity_generation
```
maryarbab, line 993: LCA activity should be: upstream-fuel_combustion as it is a diesel vehicle not electric

OpenEmissionFactorsDB.csv
```diff
@@ -990,8 +990,8 @@ Transport,Rail Transport,freight_train-route_type_domestic-fuel_source_electrici
...
+Transport,Rail Travel,passenger_train-route_type_long_distance-fuel_source_diesel,Diesel long-distance passenger train,kg-CO2e/pkm,0.044,,UBA,2020,DE,2021/09/15,Emission intensity of diesel long-distance passenger train including upstream and fuel combustion. Data from the Öko-Institut.,https://www.probas.umweltbundesamt.de/php/prozessdetails.php?id={968DDFF4-52FF-4C77-8F1D-F21DD202C11C},electricity_generation
```
maryarbab, line 994: LCA activity should be: upstream-fuel_combustion

OpenEmissionFactorsDB.csv
```diff
@@ -1009,8 +1009,8 @@ Transport,Rail Travel,passenger_train-route_type_na-fuel_source_electricity,Elec
...
-Transport,Rail Travel,passenger_train-route_type_long_distance-fuel_source_electricity,Electric remote passenger train,kg-CO2e/pkm,0.00948,,UBA,2020,DE,2021/09/15,Emission intensity of long-distance electric passenger train including upstream. Data from the Öko-Institut.,https://www.probas.umweltbundesamt.de/php/prozessdetails.php?id={A4FFA0CD-2550-4435-BBDE-00C6F1A0B22F},upstream-electricity_consumption
+Transport,Rail Travel,passenger_train-route_type_local-fuel_source_electricity,Electric local passenger train,kg-CO2e/pkm,0.0546,,UBA,2020,DE,2021/09/15,Emission intensity of electric local passenger train including upstream and electricity consumption. Data from the Öko-Institut.,https://www.probas.umweltbundesamt.de/php/prozessdetails.php?id={5A001C57-5EA0-4301-8139-575EC2E58232},electricity_generation
```
maryarbab, line 1012: Suggestion for LCA activity: upstream-electricity_consumption (if consumption is mentioned by the source) or upstream-electricity_generation (if generation is mentioned by the source)

OpenEmissionFactorsDB.csv
```diff
@@ -1009,8 +1009,8 @@ Transport,Rail Travel,passenger_train-route_type_na-fuel_source_electricity,Elec
...
+Transport,Rail Travel,passenger_train-route_type_long_distance-fuel_source_electricity,Electric long-distance passenger train,kg-CO2e/pkm,0.00948,,UBA,2020,DE,2021/09/15,Emission intensity of electric long-distance passenger train including upstream and electricity consumption. Data from the Öko-Institut.,https://www.probas.umweltbundesamt.de/php/prozessdetails.php?id={A4FFA0CD-2550-4435-BBDE-00C6F1A0B22F},electricity_generation
```
maryarbab, line 1013: Suggestion for LCA activity: upstream-electricity_consumption (if consumption is mentioned by the source) or upstream-electricity_generation (if generation is mentioned by the source)

maryarbab (changes requested):

maryarbab (approved):

maryarbab (approved):

OpenEmissionFactorsDB.csv
AnastasiaLobanova: Line 993 here - wasn't it that discussion we had about Nahverkehr meaning in German, and we came to a conclusion it means urban trains, e.g. underground trains/trams? 
https://de.wikipedia.org/wiki/Nahverkehr

PR #97, opened 2021-12-10
LCIA of electricity generation sources (UNECE)

greenercode:

- [x] New

Electricity

**What?**

Includes LCIA of electricity generation sources (source UNECE) for region EU28.

**Why?**

This source is essential in the analysis of life-cycle greenhouse gas emissions for various electricity generation sources.


OpenEmissionFactorsDB.csv
```diff
@@ -352,12 +354,32 @@ Energy,Electricity,electricity-energy_source_grid_mix,Grid mix,kg-CO2e/kWh,0.555
...
+Energy,Electricity,electricity-energy_source_natural_gas_ngcc_without_ccs,Natural gas NGCC (without CCS),kg-CO2e/kWh,0.43,,UNECE,2020,EU,2021/12/10,Emission intensity for natural gas NGCC (without CCS) for region EUR (Europe EU28). Data provided by the United Nations Economic Commission for Europe. ,https://unece.org/sed/documents/2021/10/reports/life-cycle-assessment-electricity-generation-options,lifecycle_impact_assesement
+Energy,Electricity,electricity-energy_source_nuclear_average,Nuclear average,kg-CO2e/kWh,0.0051,,UNECE,2020,EU,2021/12/10,Emission intensity for nuclear (average) for region EUR (Europe EU28). Data provided by the United Nations Economic Commission for Europe. ,https://unece.org/sed/documents/2021/10/reports/life-cycle-assessment-electricity-generation-options,lifecycle_impact_assesement
```
maryarbab, line 367: I would remove "average" here.

OpenEmissionFactorsDB.csv
```diff
@@ -75,6 +75,8 @@ Consumer Goods and Services,Vehicle Maintenance and Services,consumer_goods-type
...
 Energy,Electricity,electricity-energy_source_coal_fired_plant,Coal fired plant,kg-CO2e/kWh,1.06,10,ADEME,2018,FR,2021/08/20,Emission intensity for coal fired plant. The data scope includes upstream and combustion emissions. Updated December 2018. Published by ADEME.,https://www.bilans-ges.ademe.fr/en/accueil,upstream-fuel_combustion
+Energy,Electricity,electricity-energy_source_csp_tower,CSP tower,kg-CO2e/kWh,0.022,,UNECE,2020,EU,2021/12/10,Emission intensity for CSP tower for region EUR (Europe EU28). Data provided by the United Nations Economic Commission for Europe. ,https://unece.org/sed/documents/2021/10/reports/life-cycle-assessment-electricity-generation-options,lifecycle_impact_assesement
```
maryarbab, line 78: Do we need "Data provided by the United Nations Economic Commission for Europe." in the description as it is already noted in the source field?

maryarbab: "lifecycle_impact_assesement" is not LCA activity. It is a methodology. Does this mean that the full life cycle is considered in their analysis?

OpenEmissionFactorsDB.csv
```diff
@@ -270,7 +272,7 @@ Energy,Electricity,electricity-energy_source_grid_mix,Grid mix,kg-CO2e/kWh,0.608
...
 Energy,Electricity,electricity-energy_source_grid_mix,Grid mix,kg-CO2e/kWh,0.72328,,AIB,2019,EE,2021/07/30,Emission intensity of grid mix as reported. Published by the Association of Issuing Bodies (AIB) 2019. Includes production fuel mix factor.,https://www.aib-net.org/,electricity_generation
-Energy,Electricity,electricity-energy_source_dkis_grid_mix,Grid mix,kg-CO2e/kWh,0.55,,DISER,2018,AU-NT,2021/11/30,Emission intensity for the Darwin Katherine Interconnected System (DKIS) grid mix. Published by the Australian Government. Includes electricity generation factor (only). ,https://www.industry.gov.au/data-and-publications/national-greenhouse-accounts-factors-2019,electricity_generation
+Energy,Electricity,electricity-energy_source_dkis_grid_mix,Grid mix,kg-CO2e/kWh,0.55,,DISER,2018,AU-NT,2021/11/30,Emission intensity for the Darwin Katherine Interconnected System (DKIS) grid mix. Published by the Australian Government. Includes electricity generation factor (only).,https://www.industry.gov.au/data-and-publications/national-greenhouse-accounts-factors-2019,electricity_generation
```
maryarbab, line 275: Is "dkis" needed in the sub ID?

maryarbab (changes requested):

OpenEmissionFactorsDB.csv
```diff
@@ -75,6 +75,8 @@ Consumer Goods and Services,Vehicle Maintenance and Services,consumer_goods-type
...
 Energy,Electricity,electricity-energy_source_coal_fired_plant,Coal fired plant,kg-CO2e/kWh,1.06,10,ADEME,2018,FR,2021/08/20,Emission intensity for coal fired plant. The data scope includes upstream and combustion emissions. Updated December 2018. Published by ADEME.,https://www.bilans-ges.ademe.fr/en/accueil,upstream-fuel_combustion
+Energy,Electricity,electricity-energy_source_csp_tower,CSP tower,kg-CO2e/kWh,0.022,,UNECE,2020,EU,2021/12/10,Emission intensity for CSP tower for region EUR (Europe EU28). Data provided by the United Nations Economic Commission for Europe. ,https://unece.org/sed/documents/2021/10/reports/life-cycle-assessment-electricity-generation-options,lifecycle_impact_assesement
+Energy,Electricity,electricity-energy_source_csp_trough,CSP trough,kg-CO2e/kWh,0.042,,UNECE,2020,EU,2021/12/10,Emission intensity for CSP trough for region EUR (Europe EU28). Data provided by the United Nations Economic Commission for Europe. ,https://unece.org/sed/documents/2021/10/reports/life-cycle-assessment-electricity-generation-options,lifecycle_impact_assesement
```
maryarbab, line 79: The link source contains "LCA-2.pdf". Does this doc contain the data? Just to make sure we are providing the link to the actual data.

maryarbab (changes requested):

OpenEmissionFactorsDB.csv
greenercode: Yes, this is my assumption. Although, **life cycle impact assessment (LCIA) is the method for converting inventory data from a life cycle assessment into a set of potential impacts. This enables practitioners and decision makers to better understand the damage caused by resource use and emissions.** The current format was only indicative, let's try to figure out the best option. I will look into the methodology. :)

OpenEmissionFactorsDB.csv
greenercode: Yes!

OpenEmissionFactorsDB.csv
greenercode: Not the same PR - line has been moved here since merging, this is why it is considered as change.

OpenEmissionFactorsDB.csv
patstirling: From what I can see I suspect that this is looking at the impact of generation only at the plant, not transmission & distribution. You can see from an example figure what aspects of are included, which obviously depends on the type of power generation being considered: 
image

greenercode: Screen Shot 2021-12-13 at 9 48 11 AM
This provides a bit more insight into the LCA assessment of these (page 10).

PR #98, opened 2021-12-13
Rearrange categories alphabetically + tons ranges

greenercode:

- [x] Update

ID structure guidance 

**What?**

Rearrange categories alphabetically + correct ID structure guidance for tons in weight ranges (road transport). 

**Why?**

Improved clarity.


patstirling (changes requested):
I don't think we need to do this - there is a bigger change required to accommodate the large number of categories added with spend-based EFs, which will need some thinking. To be reviewed by the team before next steps decided

PR #103, opened 2021-12-16
Include business travel (land) BEIS

greenercode:

- [x] New

Vehicle 

**What?**

Include a set of remaining BEV, hybrid and PHEV emission factors under the business travel (land) category. Available in BEIS Greenhouse Gas Reporting: Conversion Factors 2021. **BEIS business travel only!**

**Why?**

Include business travel emission factors.


OpenEmissionFactorsDB.csv
```diff
@@ -1284,21 +1284,34 @@ Transport,Vehicle,passenger_vehicle-vehicle_type_bus-fuel_source_bev-engine_size
...
+Transport,Vehicle,passenger_vehicle-vehicle_type_dual_purpose_4x4_car-fuel_source_bev-engine_size_na-vehicle_age_na-vehicle_weight_na,Battery EV car (dual-purpose 4X4) - business travel,kg-CO2e/km,0.06819,,BEIS,2021,GB,2021/12/16,Emission intensity of battery EV car (dual-purpose 4X4) - business travel. These are sport utility vehicles (SUVs) which have off-road capabilities and four-wheel drive. Examples include: Suzuki Jimny Land Rover Discovery and Defender Toyota Land Cruiser and Nissan Pathfinder. Available in BEIS Greenhouse Gas Reporting: Conversion Factors 2021.,https://www.gov.uk/government/publications/greenhouse-gas-reporting-conversion-factors-2021,fuel_combustion
```
AnastasiaLobanova, line 1290: What does "business travel" exactly mean here? Maybe elaborate here a bit? Something like: 

Battery EV car (dual-purpose 4x4) used for business purposes

And in general why the emissions of a car used for business purposes is slightly higher? If only fuel combustion is included?

AnastasiaLobanova (dismissed):

OpenEmissionFactorsDB.csv
greenercode: Thanks @AnastasiaLobanova  this one is tricky. The values were indeed missing, but the structure needs to make sense.

OpenEmissionFactorsDB.csv
```diff
@@ -1284,21 +1284,34 @@ Transport,Vehicle,passenger_vehicle-vehicle_type_bus-fuel_source_bev-engine_size
...
+Transport,Vehicle,passenger_vehicle-vehicle_type_dual_purpose_4x4_car-fuel_source_bev-engine_size_na-vehicle_age_na-vehicle_weight_na,Battery EV car (dual-purpose 4X4) - used for business purposes,kg-CO2e/km,0.06819,,BEIS,2021,GB,2021/12/16,Emission intensity of battery EV car (dual-purpose 4X4) - business travel. These are sport utility vehicles (SUVs) which have off-road capabilities and four-wheel drive. Examples include: Suzuki Jimny Land Rover Discovery and Defender Toyota Land Cruiser and Nissan Pathfinder. Available in BEIS Greenhouse Gas Reporting: Conversion Factors 2021.,https://www.gov.uk/government/publications/greenhouse-gas-reporting-conversion-factors-2021,fuel_combustion
```
maryarbab, line 1290: I also wonder why the values are different when it is a business trip?

OpenEmissionFactorsDB.csv
```diff
@@ -1284,21 +1284,34 @@ Transport,Vehicle,passenger_vehicle-vehicle_type_bus-fuel_source_bev-engine_size
...
 Transport,Vehicle,passenger_vehicle-vehicle_type_car-fuel_source_bev-engine_size_na-vehicle_age_na-vehicle_weight_na,Battery EV car (average),kg-CO2e/km,0.05031,,BEIS,2021,GB,2021/07/23,Emission intensity of electricity generation for average battery EV car. No upstream or downstream LCA activity is included. Available in BEIS Greenhouse Gas Reporting: Conversion Factors 2021.,https://www.gov.uk/government/publications/greenhouse-gas-reporting-conversion-factors-2021,electricity_generation
+Transport,Vehicle,passenger_vehicle-vehicle_type_car-fuel_source_bev-engine_size_na-vehicle_age_na-vehicle_weight_na,Battery EV car (average) - used for business purposes,kg-CO2e/km,0.05477,,BEIS,2021,GB,2021/12/16,Emission intensity of battery EV car (average) - business travel. Available in BEIS Greenhouse Gas Reporting: Conversion Factors 2021.,https://www.gov.uk/government/publications/greenhouse-gas-reporting-conversion-factors-2021,fuel_combustion
```
maryarbab, line 1287: electricity_generation should be the lca activity for BEVs

OpenEmissionFactorsDB.csv
maryarbab, line 1290: I also wonder why the values are different when it is a business trip?

maryarbab: same comment about the lca activity

OpenEmissionFactorsDB.csv
```diff
@@ -1516,22 +1533,33 @@ Transport,Vehicle,commercial_vehicle-vehicle_type_lcv-fuel_source_petrol_phev_el
...
 Transport,Vehicle,commercial_vehicle-vehicle_type_lcv-fuel_source_petrol_phev_fuel_combustion-engine_size_gt_2000cc_lt_3000cc-vehicle_age_post_2015-vehicle_weight_na,Petrol PHEV LCV (2000-3000cc) - fuel combustion post-2015,kg-CO2e/km,0.11541749,,MfE,2020,NZ,2021/09/10,Emission intensity of commercial electric vehicle. Petrol. Plug-in-hybrid. Between 2000 and 3000cc. Light commercial vehicles. No upstream or downstream LCA activity is included. Post-2015 fleet.  Published by the Ministry for the Environment of New Zealand in 2020.,https://environment.govt.nz/publications/measuring-emissions-detailed-guide-2020/,fuel_combustion
+Transport,Vehicle,passenger_vehicle-vehicle_type_car-fuel_source_phev_electricity_consumption-engine_size_na-vehicle_age_na-vehicle_weight_na,PHEV car (average) - used for business purposes,kg-CO2e/km,0.09694,,BEIS,2021,GB,2021/12/16,Emission intensity of PHEV car (average) - business travel. Available in BEIS Greenhouse Gas Reporting: Conversion Factors 2021.,https://www.gov.uk/government/publications/greenhouse-gas-reporting-conversion-factors-2021,fuel_combustion
```
maryarbab, line 1536: the name and description should be consistent with other PHEV (electricity consumption) EFs in this category. See for example row 1537. For all of these PHEVs in electric mode, lca activity should be electricity_generation (same as row 1537).

maryarbab (changes requested):

PR #104, opened 2021-12-16
Change region UNLOCODE for line 255

greenercode:

- [x] Update

Electricity 

**What?**

Change region UNLOCODE for electricity generation EF in Canada, Manitoba. 

**Why?**

The UNLOCODE was wrong.


GeeWee:
Excellent change. A++

GeeWee (approved):

PR #106, opened 2021-12-20
WTT fuel conversion factors (BEIS)

greenercode:
Include WTT upstream fuel conversion factors (BEIS).

- [x] New

Fuel 

**What?**

Include WTT upstream fuel conversion factors (BEIS). 

**Why?**

These are necessary for scope 3 calculations.


greenercode:
Should we make a distinction that these are WTT in the name field, or is upstream a sufficient and specific enough  reference in the descriptions and LCA fields? @patstirling @AnastasiaLobanova

AnastasiaLobanova (changes requested):
Looks good! 

just to delete here also "-" in the name, for consistency

OpenEmissionFactorsDB.csv
```diff
@@ -362,9 +362,13 @@ Energy,Electricity,electricity-energy_source_photovoltaic_made_in_france,Photovo
...
 Energy,Fuel,fuel_type_avgas-fuel_use_aviation,Aviation fuel (avgas),kg-CO2e/t,3192.76,,BEIS,2021,GB,2021/11/17,Emission intensity for aviation fuel (avgas). The fuels conversion factors should be used for primary fuel sources combusted at a site or in an asset owned or controlled by the reporting organisation. Available in BEIS Greenhouse Gas Reporting: Conversion Factors 2021.,https://www.gov.uk/government/publications/greenhouse-gas-reporting-conversion-factors-2021,fuel_combustion
+Energy,Fuel,fuel_type_avgas-fuel_use_aviation_wtt,Aviation fuel (avgas) WTT,kg-CO2e/L,0.59512,,BEIS,2021,GB,2021/12/20,"Emission intensity for aviation fuel (avgas). Well-to-Tank (WTT) fuels conversion factors should be used to account for the upstream Scope 3 emissions associated with extraction, refining and transportation of the raw fuel sources to an organisation’s site (or asset), prior to combustion. Available in BEIS Greenhouse Gas Reporting: Conversion Factors 2021.",https://www.gov.uk/government/publications/greenhouse-gas-reporting-conversion-factors-2021,Well-to-Tank
```
maryarbab, line 365: I don't think we need wtt in the ID as it is the LCA activity.

maryarbab: Can you please use "Retrieved from the Conversion Factors 2021: flat file published by the UK BEIS/Defra at the source URL" for all the referencing in the descriptions across your PRs?

maryarbab (changes requested):

OpenEmissionFactorsDB.csv
AnastasiaLobanova: also the Well-to-Tank in the lca_activity field can be uncapitalised

greenercode: This is already the case @AnastasiaLobanova :)

maryarbab (approved):

AnastasiaLobanova (approved):

patstirling (changes requested):
commas in all description fields

PR #111, opened 2021-12-21
Include WTT- UK & overseas elec. (BEIS)

greenercode:
Include WTT- UK & overseas elec. (BEIS)

- [x] New

Electricity 

**What?**

Includes WTT- UK & overseas elec. (BEIS).

**Why?**

These are essential in reporting to calculate total emissions generated.


OpenEmissionFactorsDB.csv
```diff
@@ -359,130 +359,130 @@ Energy,Electricity,electricity-energy_source_onshore_wind_turbine,Onshore wind t
...
+Energy,Electricity,electricity-energy_source_grid_mix-type_wtt_generation,"WTT- Electricity generation - Hong Kong, China",kg-CO2e/kWh,0.17198,,BEIS,2021,HK,2021/12/21,"Emission intensity of WTT- Electricity generation - Hong Kong, China. Well-to-tank (WTT) conversion factors for UK and overseas electricity should be used to report the Scope 3 emissions of extraction, refining and transportation of primary fuels before their use in the generation of electricity. It should be noted that electricity generation, transmission and distribution losses have separate WTT emissions assigned to them. Available in BEIS Greenhouse Gas Reporting: Conversion Factors 2021. ",https://www.gov.uk/government/publications/greenhouse-gas-reporting-conversion-factors-2021,well_to_tank
```
AnastasiaLobanova: I suggest and instead of "," in the name column

AnastasiaLobanova (changes requested):
Nice work Stephane!! A lot of new data, really cool stuff!

I have two general suggestions, in the description instead of "," use "/" to avoid the appearance of "" and to delete  "-" in the names column, e.g.

WTT- Electricity generation - Gibraltar -> WTT electricity generation Gibraltar

OpenEmissionFactorsDB.csv
greenercode: Done - better now? 🤔

OpenEmissionFactorsDB.csv
```diff
@@ -359,130 +359,130 @@ Energy,Electricity,electricity-energy_source_onshore_wind_turbine,Onshore wind t
...
+Energy,Electricity,electricity-energy_source_grid_mix-type_wtt_td,WTT Transmission & Distribution NonOECD Europe and Eurasia (average),kg-CO2e/kWh,0.013588,,BEIS,2021,EURASIA,2021/12/21,Emission intensity of WTT Transmission & distribution NonOECD Europe and Eurasia (average). Well-to-Tank (WTT) conversion factors for UK and overseas electricity should be used to report the Scope 3 emissions of extraction/ refining and transportation of primary fuels before their use in the generation of electricity. It should be noted that electricity generation/ transmission and distribution losses have separate WTT emissions assigned to them. Available in BEIS Greenhouse Gas Reporting: Conversion Factors 2021. ,https://www.gov.uk/government/publications/greenhouse-gas-reporting-conversion-factors-2021,Well-to-Tank
```
AnastasiaLobanova: NonOECD -> Non OECD

AnastasiaLobanova (changes requested):
one small bug

OpenEmissionFactorsDB.csv
```diff
@@ -359,6 +359,130 @@ Energy,Electricity,electricity-energy_source_onshore_wind_turbine,Onshore wind t
...
+Energy,Electricity,electricity-energy_source_grid_mix-type_wtt_generation,WTT Electricity generation Africa (average),kg-CO2e/kWh,0.118733,,BEIS,2021,AF,2021/12/21,Emission intensity of WTT Electricity generation Africa (average). Well-to-Tank (WTT) conversion factors for UK and overseas electricity should be used to report the Scope 3 emissions of extraction/ refining and transportation of primary fuels before their use in the generation of electricity. It should be noted that electricity generation/ transmission and distribution losses have separate WTT emissions assigned to them. Available in BEIS Greenhouse Gas Reporting: Conversion Factors 2021. ,https://www.gov.uk/government/publications/greenhouse-gas-reporting-conversion-factors-2021,Well-to-Tank
```
maryarbab: We should not include wtt in the ID. It refers to the lca activity. We also need to be consistent with other IDs in the same category. Suggestion for the ID: electricity-energy_source_grid_mix

maryarbab: Please update the description to be: Available in BEISs Greenhouse Gas Reporting: Conversion Factors 2021-flat file.

maryarbab: Where does the fuel, which is needed to generate electricity produced? Africa or UK?

maryarbab: We need to distinguish between AF and UK here. Same as other rows.

maryarbab: lca activity needs to be in lowe case.

maryarbab (changes requested):

OpenEmissionFactorsDB.csv
greenercode: This is **WTT- UK & overseas electricity** - would you mind providing an example of how this should be distinguished? Thanks!

> We need to distinguish between AF and UK here. Same as other rows.

maryarbab (approved):

AnastasiaLobanova (approved):

PR #115, opened 2022-01-12
DISER source update

maryarbab:

- Update

_Electricity_

**What?**

**Why?**

_To link to the most recent primary source, which is DISER 2021_


AnastasiaLobanova (approved):
Looks good! Just a question about DISER - were the numbers from 2018 wrong? Thank you Maryam!

maryarbab:
> Looks good! Just a question about DISER - were the numbers from 2018 wrong? Thank you Maryam!

Nope, they were just outdated.

patstirling (approved):

OpenEmissionFactorsDB.csv
```diff
@@ -1095,8 +1095,8 @@ Materials,Mined Materials,mined_materials-type_copper_nickel_lead_zinc,Copper/ni
...
-Materials,Mining ,mining-type_well_drilling,Well drilling,kg-CO2e/USD,0.543,,EPA,2020,US,2021/10/12,Emission intensity of supply chain (with margins i.e. cradle to shelf) in US dollars spend on: well drilling. This factor is representative of the described commodity (equivalent to a goods or services category) and was calculated from 2016 data applying IPCC 4th Assessment Report conversion factors.,https://edg.epa.gov/metadata/catalog/search/resource/details.page?uuid=https://doi.org/10.23719/1517796,cradle_to_shelf
+Materials,Mining,mining-type_other_support_activities_for_mining,Other support activities for mining,kg-CO2e/USD,0.613,,EPA,2020,US,2021/10/12,Emission intensity of supply chain (with margins i.e. cradle to shelf) in US dollars spend on: other support activities for mining. This factor is representative of the described commodity (equivalent to a goods or services category) and was calculated from 2016 data applying IPCC 4th Assessment Report conversion factors.,https://edg.epa.gov/metadata/catalog/search/resource/details.page?uuid=https://doi.org/10.23719/1517796,cradle_to_shelf
```
patstirling, line 1098: This and the next two changes will cause conflicts

patstirling (changes requested):
The wording here uses "consumption" which seems right. A few non-AU DISER-related changes in this PR will cause conflicts.

OpenEmissionFactorsDB.csv
maryarbab: I am going to change it in the last PR. Should I wait to merge the last PR before merging this one then?

patstirling: I'd remove the changes from each PR to be sure

maryarbab: Done;)

OpenEmissionFactorsDB.csv
```diff
@@ -218,7 +218,7 @@ Energy,Electricity,electricity-energy_source_electricity,Electricity,kg-CO2e/USD
...
-Energy,Electricity,electricity-energy_source_dkis_grid_mix,Grid mix,kg-CO2e/kWh,0.55,,DISER,2018,AU-NT,2021/11/30,Emission intensity for the Darwin Katherine Interconnected System (DKIS) grid mix. Published by the Australian Government. Includes electricity generation factor (only).,https://www.industry.gov.au/data-and-publications/national-greenhouse-accounts-factors-2019,electricity_generation
+Energy,Electricity,electricity-energy_source_dkis_grid_mix,Grid mix,kg-CO2e/kWh,0.54,,DISER,2021,AU-NT,2021/11/30,Emission intensity for the Darwin Katherine Interconnected System (DKIS) grid mix. Published by the Australian Government 2021. Indirect (scope 2) emission factors for consumption of purchased electricity or loss of electricity from the grid.,https://www.industry.gov.au/data-and-publications/national-greenhouse-accounts-factors-2021,electricity_generation
```
patstirling, line 221: Just checking in light of other reviews: does "Indirect (scope 2) emission factors for consumption of purchased electricity or loss of electricity from the grid" imply that this is the emissions generated at consumption? If so should we change lca_activity for these EFs?

patstirling (changes requested):

OpenEmissionFactorsDB.csv
maryarbab: I know it is a bit confusing. I actually checked the source when I was updating it and the source has this explanation: 

"The estimated electricity emission factors have been aligned with the definitions used in The
Greenhouse Gas Protocol: A Corporate Accounting and Reporting Standard of the World
Resources Institute/World Business Council for Sustainable Development (the GHG
Protocol). The emission factor for scope 2 is defined in terms of energy sent out on the grid
rather than energy delivered because this effectively ensures that end users of electricity are
allocated only the scope 2 emissions attributable to the electricity they consume and not the
scope 2 emissions attributable to electricity lost in transmission and distribution."

patstirling: This to me _largely_ suggests that it is for measurement of consumed electricity, to be attributed as scope 2 by the consumer - so suggest we treat this as a consumption metric

maryarbab: Alright, changed it.

patstirling: This also maps neatly on to "use_phase" as a macro LCA category. Approved!

patstirling (approved):

PR #116, opened 2022-01-13
DEWA source update

maryarbab:

- [ ] Update

_Electricity_

**What?**

**Why?**

_To link to the most recent primary source, which is DEWA 2020_


greenercode (dismissed):

AnastasiaLobanova (approved):

OpenEmissionFactorsDB.csv
```diff
@@ -482,7 +482,7 @@ Energy,Electricity,electricity-energy_source_grid_mix,Grid mix,kg-CO2e/kWh,0.348
...
 Energy,Electricity,electricity-energy_source_grid_mix,Grid mix,kg-CO2e/kWh,0.42988,,EPA,2018,US-FL,2021/07/23,Emission intensity of grid mix as reported. Retrieved from EPA eGrid 2020. Includes generation factor.,https://www.epa.gov/egrid,electricity_generation
-Energy,Electricity,electricity-energy_source_grid_mix,Grid mix,kg-CO2e/kWh,0.4258,,DEWA,2018,AE,2021/07/30,Emission intensity of grid mix as reported. Published by Dubai Electricity & Water Authority 2018. Includes generation factor only.,https://www.dewa.gov.ae/~/media/Files/Customer/Sustainability%20Reports/DEWA%20Sustainability%20Report%202018_EN.ashx,electricity_generation
+Energy,Electricity,electricity-energy_source_grid_mix,Grid mix,kg-CO2e/kWh,0.4041,,DEWA,2020,AE,2021/07/30,Emission intensity of electricity generated as reported. Published by Dubai Electricity & Water Authority 2020. ,https://www.dewa.gov.ae/en/consumer/Sustainability/sustainability-reports,electricity_generation
```
patstirling, line 485: should this be "electricity generated" or "electricity consumed"? I think these EFs are designed to be used by consumers of the energy?

I also think we should be more explicit, say "electricity consumed in the United Arab Emirates." There's a question here about whether we should be using the term "Average" somewhere - e.g. "Average emissions generated by consumption of a unit of electricity in the United Arab Emirates" - but I'll post that for a broader discussion.

OpenEmissionFactorsDB.csv
```diff
@@ -1095,8 +1095,8 @@ Materials,Mined Materials,mined_materials-type_copper_nickel_lead_zinc,Copper/ni
...
-Materials,Mining ,mining-type_well_drilling,Well drilling,kg-CO2e/USD,0.543,,EPA,2020,US,2021/10/12,Emission intensity of supply chain (with margins i.e. cradle to shelf) in US dollars spend on: well drilling. This factor is representative of the described commodity (equivalent to a goods or services category) and was calculated from 2016 data applying IPCC 4th Assessment Report conversion factors.,https://edg.epa.gov/metadata/catalog/search/resource/details.page?uuid=https://doi.org/10.23719/1517796,cradle_to_shelf
+Materials,Mining,mining-type_other_support_activities_for_mining,Other support activities for mining,kg-CO2e/USD,0.613,,EPA,2020,US,2021/10/12,Emission intensity of supply chain (with margins i.e. cradle to shelf) in US dollars spend on: other support activities for mining. This factor is representative of the described commodity (equivalent to a goods or services category) and was calculated from 2016 data applying IPCC 4th Assessment Report conversion factors.,https://edg.epa.gov/metadata/catalog/search/resource/details.page?uuid=https://doi.org/10.23719/1517796,cradle_to_shelf
```
patstirling, line 1098: I think this is going to cause conflicts with all the other EFs - it needs to be fixed, but suggest it be fixed in the very last of this volley of PRs?

patstirling (changes requested):

OpenEmissionFactorsDB.csv
maryarbab: Yes, let's do that.

OpenEmissionFactorsDB.csv
maryarbab: The source says: Carbon emission intensity, tCO2e/MWH of electricity generated. So, I think we should also keep it as generated indicating that it does not include t&d losses.

patstirling (changes requested):

OpenEmissionFactorsDB.csv
maryarbab: Revised;)

patstirling (approved):

AnastasiaLobanova (approved):

PR #117, opened 2022-01-13
BEIS source update

maryarbab:

- Update

_All BEIS categories_

**What?**

**Why?**

_To link to the most recent primary source, which is BEIS "Conversion factors 2021: flat file"_


AnastasiaLobanova (approved):
Great stuff Maryam!

OpenEmissionFactorsDB.csv
```diff
@@ -259,7 +259,7 @@ Energy,Electricity,electricity-energy_source_grid_mix,Grid mix,kg-CO2e/kWh,0.016
...
+Energy,Electricity,electricity-energy_source_grid_mix,Grid mix,kg-CO2e/kWh,0.21233,,BEIS,2021,GB,2021/07/23,Emission intensity of electricity generated as reported. Available in BEISs Greenhouse Gas Reporting: Conversion Factors 2021-flat file. These factors include only direct emissions at UK power stations and from autogenerators plus those from the proportion of imported electricity. They do not include emissions resulting from production and delivery of fuel to these power stations.,https://www.gov.uk/government/publications/greenhouse-gas-reporting-conversion-factors-2021,electricity_generation
```
patstirling: can we replace "-flat" with ": flat" as per the text at the source URL

OpenEmissionFactorsDB.csv
```diff
@@ -482,7 +482,7 @@ Energy,Electricity,electricity-energy_source_grid_mix,Grid mix,kg-CO2e/kWh,0.348
...
 Energy,Electricity,electricity-energy_source_grid_mix,Grid mix,kg-CO2e/kWh,0.42988,,EPA,2018,US-FL,2021/07/23,Emission intensity of grid mix as reported. Retrieved from EPA eGrid 2020. Includes generation factor.,https://www.epa.gov/egrid,electricity_generation
-Energy,Electricity,electricity-energy_source_grid_mix,Grid mix,kg-CO2e/kWh,0.4258,,DEWA,2018,AE,2021/07/30,Emission intensity of grid mix as reported. Published by Dubai Electricity & Water Authority 2018. Includes generation factor only.,https://www.dewa.gov.ae/~/media/Files/Customer/Sustainability%20Reports/DEWA%20Sustainability%20Report%202018_EN.ashx,electricity_generation
+Energy,Electricity,electricity-energy_source_grid_mix,Grid mix,kg-CO2e/kWh,0.4041,,DEWA,2020,AE,2021/07/30,Emission intensity of electricity generated as reported. Published by Dubai Electricity & Water Authority 2020.,https://www.dewa.gov.ae/en/consumer/Sustainability/sustainability-reports,electricity_generation
```
patstirling: I think this may have sneaked in from another PR?

patstirling (changes requested):

OpenEmissionFactorsDB.csv
maryarbab: I intentionally changed it to dash since it would be "Available in BEISs Greenhouse Gas Reporting: Conversion Factors 2021: flat file" in some case. If it seems okay, I can change it all to ": flat".

OpenEmissionFactorsDB.csv
maryarbab: Removed it:)

OpenEmissionFactorsDB.csv
patstirling: Hm, tricky. Ok I've had a look at the site - suggest the following "Retrieved from the Conversion Factors 2021: flat file published by the UK BEIS/Defra at the source URL"

maryarbab: Done.

patstirling (dismissed):

AnastasiaLobanova (dismissed):

greenercode (approved):

grindcode (approved):

PR #120, opened 2022-01-13
ADEME source update

maryarbab:
Is this a new addition or an update to an existing factor?

Update
What category or categories does it affect?

All BEIS categories

What?

Updates to the year, source link, and description fields.

Why?

To link to the most recent primary source, which is ADEME Base Carbone - Data V21.0.csv.


greenercode (approved):

OpenEmissionFactorsDB.csv
```diff
@@ -219,143 +219,143 @@ Energy,Electricity,electricity-energy_source_na,Electricity,kg-CO2e/GBP,2.249,,B
...
-Energy,Electricity,electricity-energy_source_grid_mix,Grid mix,kg-CO2e/kWh,0.36,10,ADEME,2013,DK,2021/08/13,Emission intensity of grid mix as reported. The values take electrical and thermal kWh supply into account. Cross-border power exchanges are not included. The publication covers only direct emissions from power plants. Provided by the IEA in 2013. Published by ADEME.,https://bilans-ges.ademe.fr/en/basecarbone/donnees-consulter/liste-element/categorie/68,electricity_generation
+Energy,Electricity,electricity-energy_source_grid_mix,Grid mix,kg-CO2e/kWh,0.842,10,ADEME,2013,KW,2021/08/13,Emission intensity of grid mix as reported. The values take electrical and thermal kWh supply into account. Cross-border power exchanges are not included. The publication covers only direct emissions from power plants. Provided by the IEA in 2013. Published by ADEME. The value is in CO2.,https://bilans-ges.ademe.fr/en/basecarbone/donnees-consulter/liste-element/categorie/68,electricity_generation
```
patstirling: I think for this temporary fix this should be clearer, e.g. "NOTE: this value is provided in CO2 only and does no include the impact of other GHGs. This factor will be deprecated."

OpenEmissionFactorsDB.csv
```diff
@@ -219,143 +219,143 @@ Energy,Electricity,electricity-energy_source_na,Electricity,kg-CO2e/GBP,2.249,,B
...
+Energy,Electricity,electricity-energy_source_grid_mix,Grid mix,kg-CO2e/kWh,0.0296,10,ADEME,2013,SE,2021/08/13,Emission intensity of grid mix as reported. The values take electrical and thermal kWh supply into account. Cross-border power exchanges are not included. The publication covers only direct emissions from power plants. Provided by the IEA in 2013. Published by ADEME. The value is in CO2. The value is in CO2.,https://bilans-ges.ademe.fr/en/basecarbone/donnees-consulter/liste-element/categorie/68,electricity_generation
```
patstirling: addition made twice here

patstirling (changes requested):

OpenEmissionFactorsDB.csv
maryarbab: Revised!

patstirling: Oops - really sorry there was a typo in my text should be "does not include" not "does no include" (which sounds ok in a Scots accent but otherwise is definitely wrong)

patstirling (changes requested):

patstirling (dismissed):

AnastasiaLobanova (approved):

PR #126, opened 2022-01-17
CCF unit update

maryarbab:

- [ ] Update

_CCF_

**What?**

_Update Memory and Storage including the time constrain_

**Why?**

_To be consistent with the methodology_


GeeWee (dismissed):
Matches my understanding, but I'll defer to Pat or Isis for the final judgement. @Isis note that I don't think have a unit for `gb-hour` in the API currently.

grindcode:
@GeeWee you are right, we will need to add that unit. Question is if this turns into a breaking change or defaults will handle the cases where no time is defined for storage.

grindcode (approved):

GeeWee:
@grindcode it will be a breaking change, but I think we can consider it a bugfix - as the old numbers are _wrong_

GeeWee (approved):
new lca changes look good to me!

PR #133, opened 2022-01-21
updating the UBA source (with the latest main)

AnastasiaLobanova:

- [ ] New
- [x] Update

Multiple categories from the UBA source

**What?**

Update the source links and some other small errors

**Why?**

Data Health support


OpenEmissionFactorsDB.csv
```diff
@@ -1511,46 +1511,46 @@ Transport,Road Transport,freight_vehicle-vehicle_type_light_delivery-fuel_source
...
+Transport,Road Travel,passenger_vehicle-vehicle_type_car-fuel_source_bev-distance_na-engine_size_na,Electric passenger car,kg-CO2e/pkm,0.0621,,UBA,2020,DE,2021/09/15,Emission intensity of electric passenger car medium. Direct electricity consumption and upstream LCA activity is included. ,https://www.probas.umweltbundesamt.de/php/prozessdetails.php?id={D8D60864-21D9-42A2-9E56-2C1AD4D9CBC5},upstream-electricity_consumption
+Transport,Road Travel,passenger_vehicle-vehicle_type_bus_line-fuel_source_fcev-distance_na-engine_size_na,Hydrogen bus urban,kg-CO2e/pkm,0.0251,,UBA,2020,DE,2021/09/15,Emission intensity of hydrogen urban bus. Upstream and fuel combustion LCA activity is included. ,https://www.probas.umweltbundesamt.de/php/prozessdetails.php?id={15F06C28-861B-49E5-9BE9-B761769CAD27},upstream-fuel_combustion
```
patstirling, line 1550: Name format here is typically "Hydrogen bus (urban)"

OpenEmissionFactorsDB.csv
```diff
@@ -1511,46 +1511,46 @@ Transport,Road Transport,freight_vehicle-vehicle_type_light_delivery-fuel_source
...
-Transport,Road Travel,passenger_vehicle-vehicle_type_car-fuel_source_fcev-distance_na-engine_size_small,Hydrogen passenger car small,kg-CO2e/pkm,0.0808,,UBA,2020,DE,2021/09/15,Emission intensity of hydrogen passenger car small. Data from the Öko-Institut. Value applicable to 2020. Provided by the UBA.,https://www.probas.umweltbundesamt.de/php/prozessdetails.php?id={AA2230CD-0099-4BB9-BD7A-90E1268A61A5},unknown
+Transport,Road Travel,passenger_vehicle-vehicle_type_car-fuel_source_bev-distance_na-engine_size_na,Electric passenger car,kg-CO2e/pkm,0.0621,,UBA,2020,DE,2021/09/15,Emission intensity of electric passenger car medium. Direct electricity consumption and upstream LCA activity is included. ,https://www.probas.umweltbundesamt.de/php/prozessdetails.php?id={D8D60864-21D9-42A2-9E56-2C1AD4D9CBC5},upstream-electricity_consumption
```
patstirling, line 1549: Where size of car is provided we usually put this in the ID as "vehicle_type_medium_car" and in the name as "Battery EV car (medium)" - unless this is meant to be an "average" EV car in which case we should simply say "vehicle_type_car" and "Battery EV car"

This applies to the name and ID for several following EFs in the vehicle category - seek to conform with other EFs in the DB

maryarbab (approved):

OpenEmissionFactorsDB.csv
```diff
@@ -1638,17 +1638,17 @@ Transport,Sea Transport,sea_freight-vessel_type_cargo-route_type_na-vessel_lengt
...
+Transport,Sea Transport,sea_freight-vessel_type_tanker_oil-route_type_na-vessel_length_suezmax-tonnage_75000t-fuel_source_bunker_c_oil,Oil tanker (Suezmax),kg-CO2e/tkm,0.00526,,UBA,2020,DE,2021/09/15,Emission intensity for 75000t oil tanker (Suezmax type). Upstream and fuel combustion LCA activity is included. ,https://www.probas.umweltbundesamt.de/php/prozessdetails.php?id={C04CE7CC-1470-4A37-BA94-E97669CC248A},upstream-fuel_combustion
+Transport,Sea Transport,sea_freight-vessel_type_cargo-route_type_international-vessel_length_na-tonnage_60000t-fuel_source_bunker_c_oil,Overseas oil ship container (medium),kg-CO2e/tkm,0.0118,,UBA,2020,DE,2021/09/15,Emission intensity for overseas average container oil ship. Upstream and fuel combustion LCA activity is included. ,https://www.probas.umweltbundesamt.de/php/prozessdetails.php?id={DADCF0F1-E502-40B2-8A25-43E9356A7919},upstream-fuel_combustion
```
patstirling, line 1651: From what I can see in the source I think this should be "Container ship (medium)" in name and description - we can drop the oil and the overseas bit, a container ship is almost never not a) going overseas or b) powered by oil!

patstirling (changes requested):

OpenEmissionFactorsDB.csv
AnastasiaLobanova: Ok!

OpenEmissionFactorsDB.csv
AnastasiaLobanova: We cannot change IDs this time. Isis said to adjust this in the next release! That's why I have reverted it to the initial version

OpenEmissionFactorsDB.csv
AnastasiaLobanova: Okay!

OpenEmissionFactorsDB.csv
AnastasiaLobanova: I suggest we create a separate PR for tackling the cars in UBA - all IDs have to be changed and the naming also

patstirling: Yep, sounds good

PR #134, opened 2022-01-24
adding BEIS biofuel WTT (latest main)

AnastasiaLobanova:

- [x] New
- [ ] Update

Fuel

**What?**

adding the BEIS well-to-tank EFs for biofuel sources 

**Why?**

important addition for calculating Scope 3 emissions


patstirling (approved):
Looks clean!!

PR #135, opened 2022-01-24
WTT fuel conversion factors (BEIS) - 2

maryarbab:
Include WTT upstream fuel conversion factors (BEIS).

Is this a new addition or an update to an existing factor?

 New

What category or categories does it affect?

Fuel

What?

Include WTT upstream fuel conversion factors (BEIS).

Why?

These are necessary for scope 3 calculations.


AnastasiaLobanova (dismissed):
Commas in the descriptions can be replaced with "and"

Seems like "well-to-tank" in the lca activity field should be changed to "well_to_tank"

PR #149, opened 2022-01-28
Fix minor issues

maryarbab:

- [*] Update

_Multiple_

**What?**

_Tackle minor issues spotted in the OEFDB including such as changing "Available in" to "Retrieved from" and updating descriptions for the CO2 only EFs._

**Why?**

_To polish the OEFDB._


AnastasiaLobanova (dismissed):

OpenEmissionFactorsDB.csv
```diff
@@ -428,153 +428,153 @@ Energy,Electricity,electricity-energy_source_grid_mix,Grid mix,kg-CO2e/kWh,0.912
...
-Energy,Electricity,electricity-energy_source_grid_mix,Grid mix,kg-CO2e/kWh,0.17103,,AIB,2020,ES,2021/07/30,Emission intensity of grid mix as reported. Published by the Association of Issuing Bodies (AIB). Available in 2020 European Residual Mix Report. Includes production fuel mix factor.The value is in CO2.,https://www.aib-net.org/facts/european-residual-mix,electricity_generation
+Energy,Electricity,electricity-energy_source_grid_mix,Grid mix,kg-CO2e/kWh,0.77669,,AIB,2020,RS,2021/07/30,Emission intensity of grid mix as reported. Published by the Association of Issuing Bodies (AIB). Retrieved from 2020 European Residual Mix Report. Includes production fuel mix factor.NOTE: this value is provided in CO2 only and does no include the impact of other GHGs. This factor will be deprecated.,https://www.aib-net.org/facts/european-residual-mix,electricity_generation
```
patstirling: "no include" -> "not include" for this and numerous other lines. We've also recently decided we will keep these factors in but not provide CO2e numbers, so we can remove "This factor will be deprecated" from this and other CO2-only EFs

patstirling (changes requested):
Could more detail be provided in the description of this PR? It's good practice to provide detail which describes at least each category of change, even if it's fairly broad - and very helpful when reviewing so you know what to look out for!

OpenEmissionFactorsDB.csv
maryarbab: Great catch! Done:)

patstirling (approved):

PR #150, opened 2022-01-31
GHG protocol-Fuel category

maryarbab:

[*] New

_Fuel_

**What?**

_Adding GHG protocol EFs along with the constituent gasses and scope as new columns._

**Why?**

_To grow the scope and coverage of the OEFDB._


patstirling (changes requested):
- Can we make the columns headers named and in the order provided in the spec (kgCO2e-AR5 > kgCO2e-AR4 > kgCO2 > kgCH4 > kgN2O > kgCO2e-OtherGHGs-AR5 > kgCO2e-OtherGHGs-AR4)
- mmBtu -> MMBTU
- Remove "Combustion of fossil fuels in vehicles (including cars/trucks/planes and boats) emits carbon dioxide/methane and nitrous oxide into the atmosphere."
- Remove "The combustion of fossil fuels by natural gas boilers/diesel generators and other equipment emits carbon dioxide/methane and nitrous oxide into the atmosphere."
- Remove "Retrieved from the GHG Emissions Calculation Tool.xlsx."

AnastasiaLobanova (dismissed):
Great work Maryam! We might need to adjust the github checker after merging this into main!

GeeWee:
We have a couple of things that rely on having the URL to the primary source in the OEFDB. I'm not well versed in the entire discussion, but I think I'd try to keep it in if possible. Anecdotally having a direct link's also made my research of some of the emission factors much, much easier.

PR #157, opened 2022-02-04
Rename id and unit fields, consolidate units

patstirling:

- [ ] New
- [X] Update

All EFs

**What?**

Rename id and unit fields, consolidate units to SI or otherwise standardise. Refactor the 15 gCO2e EFs to be kg CO2e.

**Why?**

With constituent gases being included, we need to have units of activity represented separately to units of gas emitted. This also resolves a number of egrarious unit differences throughout the DB


DATA_GUIDANCE.md
```diff
@@ -37,21 +37,23 @@ Display name of the emission factor. This field provides a human-readable versio
...
-- `kgCO2e/kWh` is kgCO2e emissions per kilowatt-hour of electricity consumed or expended.
+The emissions produced by this activity per `activity_unit`. This is the linear calculation factor used to perform emission calculations.
 
-### factor
+The following emission volumes are available (NB not every gas will be available for each activity):
 
-The emissions produced by this activity in **kgCO2e** per `unit`. This is the linear calculation factor used to perform the emission calculation.
-
-The value is always expressed in kgCO2e, also known as kgCO2 equivalent. This value includes all emitted GHG emissions and is computed based on GWP (global warming potential). 
+- `kgCO2e-AR5`: 100-year Global Warming Potential of all included GHGs expressed in kgCO2-equivalent according to the IPCC's Fifth Assessment Report
```
maryarbab, line 50: I am not sure if we can use "all" here. We are reporting kgCO2e-AR5 and kgCO2e-AR4 reported by the source and they may not have included ALL gases.

DATA_GUIDANCE.md
```diff
@@ -37,21 +37,23 @@ Display name of the emission factor. This field provides a human-readable versio
...
+- `kgCO2e-AR4`: 100-year Global Warming Potential of all included GHGs expressed in kgCO2-equivalent according to the IPCC's Fourth Assessment Report
+- `kgCO2`: Carbon dioxide emissions
+-	`kgCH4`: Methane emissions
```
maryarbab, line 53: There seems to be a space here before the kgCH4

maryarbab (changes requested):

DATA_GUIDANCE.md
patstirling: yup - that's why I've said "all included" not "all"

DATA_GUIDANCE.md
patstirling: Think this is a quirk of the delta view - it's fine in the code and in the output 
image

maryarbab (approved):

AnastasiaLobanova (approved):

PR #175, opened 2022-02-21
Install python-based validation using git instead of pip

motin:
**What?**

Install python-based validation using git instead of pip.

**Why?**

This avoids having to publish to pypi in order to update the validation functionality.


.github/workflows/validate-using-python-oefdb-sdk.yaml
```diff
@@ -0,0 +1,50 @@
...
+        uses: snok/install-poetry@v1
+        with:
+          virtualenvs-create: true
+          virtualenvs-in-project: true
+      - name: Load cached venv
+        id: cached-poetry-dependencies
+        uses: actions/cache@v2
+        with:
+          path: python-oefdb-sdk/.venv
+          key: venv-${{ runner.os }}-${{matrix.python}}-${{ hashFiles('python-oefdb-sdk/pyproject.toml') }}
```
GeeWee, line 43: shouldn't we use hashfiles on the poetry.lock file?

GeeWee (commented):
Approved with suggestions - I think there's an issue with the caching, but I don't think we need a second review pass

GeeWee (approved):

motin:
> I think there's an issue with the caching, but I don't think we need a second review pass

One annoying thing is that the cache currently only gets stored when the action runs successfully, e.g. the CSV file validates. It'd be better if the cache got stored based on the success of the install dependencies step, but I haven't check if this is possible/easy.

.github/workflows/validate-using-python-oefdb-sdk.yaml
GeeWee: Oh you're right. Let's include it.

motin:
Btw, confirmed that caching works well once the workflow passes: https://github.com/climatiq/Open-Emission-Factors-DB/runs/5272958230?check_suite_focus=true - the check takes 17s when cached:
image

PR #176, opened 2022-02-21
Restore js-based validation workflow, to run alongside the python-based validation

motin:
**What?**

Restore js-based validation workflow, to run alongside the python-based validation

**Why?**

The python-based validation is still missing some of the validations from the js-based validation workflow.


GeeWee (approved):
Works for me.

PR #178, opened 2022-02-26
Update flight regions to reflect global flights

patstirling:

- [ ] New
- [x] Update

Air travel, air freight

**What?**

Update to BEIS-issued "outside UK" flights to have GLOBAL region rather than UK. Also tidy up some names and apply consistent IDs across sources

**Why?**

In order to correctly reflect regionality of flights and provide consistent metadata in the IDs


AnastasiaLobanova (approved):
looks clean!

PR #180, opened 2022-03-02
adding gridmix data from GHG Protocol Scope 3 Calculator

AnastasiaLobanova:

- [x] New
- [ ] Update

Electricity

**What?**

adding gridmix data from GHG Protocol Scope 3 Calculator

**Why?**

Gridmix data for some countries that we do not have in the db yet


patstirling (changes requested):
Looking good! Think lca_activity should be "electricity_generation" and scope should be "2". Also, thanks for re-ordering the DB, I should have with my flights update... this is a bit cheeky and not at all to do with this PR but I noticed that one EF (line 2115) has "fight" instead of "flight", would you mind correcting this typo as a big favour??

AnastasiaLobanova:
> Looking good! Think lca_activity should be "electricity_generation" and scope should be "2". Also, thanks for re-ordering the DB, I should have with my flights update... this is a bit cheeky and not at all to do with this PR but I noticed that one EF (line 2115) has "fight" instead of "flight", would you mind correcting this typo as a big favour??

maryarbab (dismissed):

maryarbab (approved):

patstirling (approved):
Go go go!

PR #183, opened 2022-03-07
Update OpenEmissionFactorsDB.csv

patstirling:

- [x] New
- [ ] Update

Fuel

**What?**

Adding EPA 2021 Fuel EFs

**Why?**

To complete 2021 EPA EFs


OpenEmissionFactorsDB.csv
```diff
@@ -1155,31 +1232,51 @@ Energy,Fuel,fuel_type_motor_gasoline-fuel_use_gasoline_passenger_cars,Motor gaso
...
+Energy,Fuel,fuel_type_naphtha_lt_401_deg_f-fuel_use_stationary_combustion,Naphtha (<401 deg F),MMBTU,not-supplied,not-supplied,68.02,0.003,0.0006,not-supplied,not-supplied,,1,fuel_combustion,EPA,2021,US,2022/01/22,Emission intensity of stationary combustion of fuel. Retrieved from the GHG Emission Factors Hub (xlsx) file published by the US EPA at the source URL.,https://www.epa.gov/climateleadership/ghg-emission-factors-hub
+Energy,Fuel,fuel_type_naphtha_lt_401_deg_f-fuel_use_stationary_combustion,Naphtha (<401 deg F),gal (US),not-supplied,not-supplied,8.5,0.00038,0.00008,not-supplied,not-supplied,,1,fuel_combustion,EPA,2021,US,2022/01/22,Emission intensity of stationary combustion of fuel. Retrieved from the GHG Emission Factors Hub (xlsx) file published by the US EPA at the source URL.,https://www.epa.gov/climateleadership/ghg-emission-factors-hub
```
AnastasiaLobanova, line 1241: do we want to have the ids consistent with the existing ones? (see below the formatting of <401 deg F)

OpenEmissionFactorsDB.csv
```diff
@@ -1155,31 +1232,51 @@ Energy,Fuel,fuel_type_motor_gasoline-fuel_use_gasoline_passenger_cars,Motor gaso
...
+Energy,Fuel,fuel_type_other_oil_gt_401_deg_f-fuel_use_stationary_combustion,Other Oil (>401 deg F),MMBTU,not-supplied,not-supplied,76.22,0.003,0.0006,not-supplied,not-supplied,,1,fuel_combustion,EPA,2021,US,2022/01/22,Emission intensity of stationary combustion of fuel. Retrieved from the GHG Emission Factors Hub (xlsx) file published by the US EPA at the source URL.,https://www.epa.gov/climateleadership/ghg-emission-factors-hub
+Energy,Fuel,fuel_type_other_oil_gt_401_deg_f-fuel_use_stationary_combustion,Other Oil (>401 deg F),gal (US),not-supplied,not-supplied,10.59,0.00042,0.00008,not-supplied,not-supplied,,1,fuel_combustion,EPA,2021,US,2022/01/22,Emission intensity of stationary combustion of fuel. Retrieved from the GHG Emission Factors Hub (xlsx) file published by the US EPA at the source URL.,https://www.epa.gov/climateleadership/ghg-emission-factors-hub
```
AnastasiaLobanova, line 1265: same here - see below the formatting of the id

AnastasiaLobanova (changes requested):
Looks good to me! Great job, the only question is if we want to be consistent with other IDs, as mentioned above

OpenEmissionFactorsDB.csv
patstirling: Good spot! In reviewing these I've spotted some illegal characters in existing IDs. Will resolve these

AnastasiaLobanova: Hm - it should have been spotted by the Python checker! Can you point me please where are the illegal characters? (So I can add those to the checker...)

patstirling: found `.` and also capital letters (`401degF`) - both of which I thought were not allowed, but may be wrong in one or both instances...

AnastasiaLobanova: The capital letters - are an interesting question! :) dots are allowed, as far as I remember

maryarbab: Another consistency question: Shouldn't only the first letter of the the first word be capitalized in the name field?

AnastasiaLobanova (approved):

PR #188, opened 2022-03-08
Add EPA 2021 grid EFs and fix 2022 non-CO2

patstirling:

- [x] New
- [x] Update

Electricity

**What?**

Add EPA 2021 grid EFs and fix 2022 non-CO2

**Why?**

Completing 2021 EPA dataset and removing CO2e conversions from raw CH4 and NO2 numbers in 2022 dataset


maryarbab:
Looks clean! Just one comment: shouldn't we mention the subregion name in the description to be consistent with GHG protocol EFs (those from EPA)?

PR #196, opened 2022-03-16
Additional year columns - ADEME & BEIS

maryarbab:

- [*] Update

_Multiple_

**What?**

_Addition of "years_valid" and "year_calculated_from" columns_

**Why?**

_To clarify the years, which each data set has been calculated from and the years it is valid for._


AnastasiaLobanova (approved):
Looks good!

PR #209, opened 2022-03-25
adding all EXIOBASE

AnastasiaLobanova:

- [x] New
- [ ] Update

multiple

**What?**

adding all EXIOBASE

**Why?**

ingress of all EXIOBASE


maryarbab (commented):
Looks awesome! Are we planning to keepl the UK ones?

PR #215, opened 2022-04-06
Pin versions of validation-related dependencies to resolve a GitHub action caching issue

motin:
**What?**

Attempts to resolve a GitHub action caching issue.
The changes here ought to make caching work more reliably in the future. It seems that the caching feature broke because of Python releasing new minor versions. With these changes, Python-related versions are pinned down to known working ones, so that it shouldn't break on new Python releases.

Since the cache only gets updated once validation passes and validation currently fails, we need to resolve the validation errors before we can know for sure that the caching issue is resolved.

**Why?**

A poetry/python-related caching issue started cropping up recently, eg in https://github.com/climatiq/Open-Emission-Factors-DB/runs/5848156989?check_suite_focus=true

### DIFF
--- a/.github/workflows/validate-using-python-oefdb-sdk.yaml
+++ b/.github/workflows/validate-using-python-oefdb-sdk.yaml
@@ -33,6 +33,7 @@ jobs:
       - name: Install Poetry
         uses: snok/install-poetry@v1
         with:
+          version: 1.1.13
           virtualenvs-create: true
           virtualenvs-in-project: true
       - name: Load cached venv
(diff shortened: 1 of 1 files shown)


GeeWee (approved):
Seems like it'd be worth a shot. You could comment out the validation steps temporarily to see if the new cache works.

AnastasiaLobanova (approved):
thank you Fred!

motin:
> Seems like it'd be worth a shot. You could comment out the validation steps temporarily to see if the new cache works.

That would allow us to check if the caching works as long as no new Python version is deployed, but the true test will be to see how it works after the next minor Python release works. :)

GeeWee:
That's true!

PR #217, opened 2022-04-06
Adding latest sector-category-id structure file

AnastasiaLobanova:

- [ ] New
- [x] Update

Metadata

**What?**

Adding latest sector-category-id structure file

**Why?**

To improve documentation around the OEFDB


maryarbab (approved):
Looks great!

PR #222, opened 2022-04-06
BEIS ingress v3 - Accommodation

maryarbab:

- [*] Update

_Accommodation_

**What?**

_Complete BEIS Accommodation category emissions factors_

**Why?**

_To update the existing EFs and add any missing ones from the source._


AnastasiaLobanova (approved):
Looks good!

patstirling (approved):

PR #226, opened 2022-04-12
Updating Sector-Category-ID guidance

AnastasiaLobanova:

- [ ] New
- [ ] Update

None - affects only docs 

**What?**

Updating Sector-Category-ID guidance

**Why?**

Improving metadata around the OEFDB


patstirling (approved):
Looks good!!

maryarbab (approved):

PR #227, opened 2022-04-13
Update Singapore electricity generation data

digitalronin:

- [ ] New
- [x] Update

Singapore electricity generation data

**What?**

The data used for the emission factor for 
electricity generation in Singapore came from BEIS
in the UK.

The Singapore government publish their own data on
their Energy Market Authority (EMA) website at;

https://www.ema.gov.sg/

The full route to this data, for future updates;

1. Visit https://www.ema.gov.sg/
2. Click [Electricity](https://www.ema.gov.sg/Electricity.aspx)
3. Click [Browse statistics available](https://www.ema.gov.sg/Statistics.aspx)
4. Check "Emissions" and click the Search button

The only result (at time of writing) was the
source data for this change.

**Why?**

I believe the Singapore Energy Market Authority (EMA) data to be more accurate than the
previous data from BEIS in the UK because;

a) Singapore is likely to have more accurate data
about their own country's electricity generation
than BEIS, and

b) The EMA emission factor is almost 4 times
higher than the BEIS figure, which suggest that
Singapore are trying to report their emissions
honestly


patstirling (changes requested):
Thanks for the PR! Appreciate the time taken to get Singapore's government-issued grid EFs into the OEFDB. The EF you have identified to replace here is actually a "well-to-tank" factor released by the UK government for estimation of upstream impact of electricity in a given country. I suggest that rather than replacing this scope 3 factor, you add new lines to the OEFDB for Singapore scope 2 & 3 emissions. Looking at the source, I'd suggest 3 factors (e.g. rows) here, one "Electricity generation – average operating margin" (kgCO2 = 0.4080), one "Electricity generation – build margin" (kgCO2 = 0.4022) and one "Electricity generation - Upstream Fugitive Methane" (kgMH4 = 0.00212) for 2020. The difference between these should be made clear in the description field, based on the notes from the source, and in the `lca_activity` column (the value "well_to_tank" in the lca_activity column should be updated to "electricity_generation" for the generation EFs to reflect it being a scope 2 factor). CO2 values should go into the `kgCO2` column and methane into the `kgCH4` column; "not-supplied" should be placed in all GHG gas columns other than these, on each line. This includes `kgCO2e-AR4` and `kgCO2e-AR5` - these need only be populated if the source provides them, as the API will calculate these from the raw (unfactored) GHG numbers. Please note that equivalencies will not be able to be provided in this case as full CO2, CH4 and NO2 emissions need to be reported at each stage for valid calculations. The API does not yet report "incomplete" factors like these, but we are very happy to have them in the OEFDB so that we can provide them in the future (a feature we have on the roadmap).

I appreciate there's quite a lot here - very happy to help with suggestions once you have formulated the new lines!

digitalronin:
I've made the changes as I understand them.

Unfortunately, this means I won't be able to use the Climatiq API as I'd hoped. I wanted to get an easy answer to the question; "What's the kgCO2e for 100kWh of electricity in Singapore?'

The BEIS data is too inaccurate, but it sounds like I won't be able to get the EMA data from the Climatiq API, even if this PR gets merged. So, I'll have to maintain a workaround in my code.

OpenEmissionFactorsDB.csv
```diff
@@ -446,7 +446,10 @@ Energy,Electricity,electricity-energy_source_grid_mix,Electricity generation –
...
+Energy,Electricity,electricity-energy_source_grid_mix,Electricity generation – average operating margin,kWh,not-supplied,not-supplied,0.4080,not-supplied,not-supplied,not-supplied,not-supplied,,2,electricity_generation,EMA,2021,2020,,SG,,2022/04/13,Data published by the Singapore Energy Market Authority (EMA) in 2021. More information available at https://www.ema.gov.sg/singapore-energy-statistics/. The Grid Emission Factor (GEF) measures the average CO2 emission emitted per MWh of electricity. It can be calculated using the Average Operating Margin (OM) or the Build Margin (BM) method. The OM measures the system-wide emissions factor while the BM measures the emissions factor of newer facilities. More details on the GEF could be found in the "Technical Notes" section of the Singapore Energy Statistics (SES) publication.,https://www.ema.gov.sg/statistic.aspx?sta_sid=20140729MPY03nTHx2a1
```
patstirling, line 450: I've suggested:
- adding "2020" as the year_calculated
- removing the reference to "MWh" as this appears to be an error on the source's behalf (they state everything in kWh). 
- As we've added a "contributor" column to the OEFDB I've also put a link to your Git profile to credit you as the contributor.

```suggestion
Energy,Electricity,electricity-energy_source_grid_mix,Electricity generation – average operating margin,kWh,not-supplied,not-supplied,0.4080,not-supplied,not-supplied,not-supplied,not-supplied,,2,electricity_generation,EMA,2021,2020,2020,SG,,https://github.com/digitalronin,2022/04/13,Data published by the Singapore Energy Market Authority (EMA) in 2021. More information available at https://www.ema.gov.sg/singapore-energy-statistics/. The Grid Emission Factor (GEF) measures the average CO2 emission emitted per unit of electricity. It can be calculated using the Average Operating Margin (OM) or the Build Margin (BM) method. The OM measures the system-wide emissions factor while the BM measures the emissions factor of newer facilities. More details on the GEF could be found in the "Technical Notes" section of the Singapore Energy Statistics (SES) publication.,https://www.ema.gov.sg/statistic.aspx?sta_sid=20140729MPY03nTHx2a1
```

OpenEmissionFactorsDB.csv
```diff
@@ -446,7 +446,10 @@ Energy,Electricity,electricity-energy_source_grid_mix,Electricity generation –
...
+Energy,Electricity,electricity-energy_source_grid_mix,Electricity generation – build margin,kWh,not-supplied,not-supplied,0.4022,not-supplied,not-supplied,not-supplied,not-supplied,,2,electricity_generation,EMA,2021,2020,,SG,,2022/04/13,Data published by the Singapore Energy Market Authority (EMA) in 2021. More information available at https://www.ema.gov.sg/singapore-energy-statistics/. The Grid Emission Factor (GEF) measures the average CO2 emission emitted per MWh of electricity. It can be calculated using the Average Operating Margin (OM) or the Build Margin (BM) method. The OM measures the system-wide emissions factor while the BM measures the emissions factor of newer facilities. More details on the GEF could be found in the "Technical Notes" section of the Singapore Energy Statistics (SES) publication.,https://www.ema.gov.sg/statistic.aspx?sta_sid=20140729MPY03nTHx2a1
```
patstirling, line 451:
```suggestion
Energy,Electricity,electricity-energy_source_grid_mix,Electricity generation – build margin,kWh,not-supplied,not-supplied,0.4022,not-supplied,not-supplied,not-supplied,not-supplied,,2,electricity_generation,EMA,2021,2020,2020,SG,,https://github.com/digitalronin,2022/04/13,Data published by the Singapore Energy Market Authority (EMA) in 2021. More information available at https://www.ema.gov.sg/singapore-energy-statistics/. The Grid Emission Factor (GEF) measures the average CO2 emission emitted per unit of electricity. It can be calculated using the Average Operating Margin (OM) or the Build Margin (BM) method. The OM measures the system-wide emissions factor while the BM measures the emissions factor of newer facilities. More details on the GEF could be found in the "Technical Notes" section of the Singapore Energy Statistics (SES) publication.,https://www.ema.gov.sg/statistic.aspx?sta_sid=20140729MPY03nTHx2a1
```

OpenEmissionFactorsDB.csv
```diff
@@ -446,7 +446,10 @@ Energy,Electricity,electricity-energy_source_grid_mix,Electricity generation –
...
+Energy,Electricity,electricity-energy_source_grid_mix,Electricity generation - Upstream Fugitive Methane,kWh,not-supplied,not-supplied,not-supplied,0.00212,not-supplied,not-supplied,not-supplied,,3,well_to_tank,EMA,2021,2020,,SG,,2022/04/13,Data published by the Singapore Energy Market Authority (EMA) in 2021. More information available at https://www.ema.gov.sg/singapore-energy-statistics/. The Grid Emission Factor (GEF) measures the average CO2 emission emitted per MWh of electricity. It can be calculated using the Average Operating Margin (OM) or the Build Margin (BM) method. The OM measures the system-wide emissions factor while the BM measures the emissions factor of newer facilities. More details on the GEF could be found in the "Technical Notes" section of the Singapore Energy Statistics (SES) publication.,https://www.ema.gov.sg/statistic.aspx?sta_sid=20140729MPY03nTHx2a1
```
patstirling, line 452:
```suggestion
Energy,Electricity,electricity-energy_source_grid_mix,Electricity generation - Upstream Fugitive Methane,kWh,not-supplied,not-supplied,not-supplied,0.00212,not-supplied,not-supplied,not-supplied,,3,well_to_tank,EMA,2021,2020,2020,SG,,https://github.com/digitalronin,2022/04/13,Data published by the Singapore Energy Market Authority (EMA) in 2021. More information available at https://www.ema.gov.sg/singapore-energy-statistics/. The Grid Emission Factor (GEF) measures the average CO2 emission emitted per unit of electricity. It can be calculated using the Average Operating Margin (OM) or the Build Margin (BM) method. The OM measures the system-wide emissions factor while the BM measures the emissions factor of newer facilities. More details on the GEF could be found in the "Technical Notes" section of the Singapore Energy Statistics (SES) publication.,https://www.ema.gov.sg/statistic.aspx?sta_sid=20140729MPY03nTHx2a1
```

patstirling (changes requested):
Thanks for resubmitting - this looks great. I've made a couple of suggestions, including the addition of your git handle as we have added a contributor column to ensure credit is given to contributors (you are our first direct contributor!) 

Once you have made final changes I'll need to re-base against the latest main - I'm happy to do that and then commit & merge.

We'll be working on an API feature to allow "incomplete" EFs like this one to be pulled through, which will be fairly accurate, if not totally comparable to the more complete EFs produced by other agencies. I definitely would not use the BEIS data - it isn't designed to be used for complete assessment, only for supplementing EFs that don't include well-to-tank impacts (which is most of them). In fact the most accurate approach may be to use the Singapore-issued generation+upstream EFs and then add the BEIS-issued WTT numbers for both generation and transmission & distribution. This gets quite complex - which is why we are looking at providing an endpoint which combines all of these aspects for you.

OpenEmissionFactorsDB.csv
digitalronin: Thanks for getting back to me @patstirling 

I've incorporated the changes you suggested, so this should be ready for rebasing and merging.

The proposed new endpoint sounds great - can't wait to start using it!

patstirling:
PR looks great! I've just clocked that this is on a separate fork, so I can't rebase - would you be able to do that based on the latest main and I'll review & approve?

digitalronin:
@patstirling I've raised a new PR - #232 - so this one can be closed.

PR #241, opened 2022-05-05
EPA 2022 ingress: waste and refrigerants sectors

AnastasiaLobanova:

- [x] New
- [ ] Update

Waste and Refrigerants sectors

**What?**

EPA 2022 ingress: waste and refrigerants sectors

**Why?**

Annual update of EFs from primary sources


maryarbab (approved):
Looks great!

PR #245, opened 2022-05-11
adding data quality flags to EXIOBASE DE GB and US data

AnastasiaLobanova:

- [ ] New
- [x] Update

Multiple

**What?**

adding data quality flags to EXIOBASE DE GB and US data

**Why?**

Data quality revision


patstirling (approved):
Looks rock-solid to me! Only thing I noticed was that the various DE landfill EFs seemed very high, but if they are each part of their respective clusters (e.g. other `Paper for treatment: landfill` values in the EXIOBASE dataset are around 34.5 as well) then all good.

PR #248, opened 2022-05-13
Retire old GitHub checkers

igmoweb:
Just retire old GH Actions chekers that aren't being used anymore

@motin, one of these is yours, can you confirm that we can retired for the good? Thanks


grindcode (approved):

GeeWee:
BREAKING THE LAW

PR #252, opened 2022-05-19
adding EXIOBASE for CH CN CA and BE

AnastasiaLobanova:

- [x] New
- [ ] Update

Multiple

**What?**

adding EXIOBASE for CH CN CA and BE

**Why?**

Part of EXIOBASE ingress


patstirling (changes requested):
Looks good! There is one interesting outlier, which is `electricity-energy_source_coal` for CA - it's marked as an outlier at 16 but is lower than BG at 20, which is not?

I also noticed that the AT waste_management-type_food_waste_for_treatment_landfill (from the last PR) is marked as an outlier at 30 while AU and BG are not at 37?

fuel_type_motor_fuel-fuel_use_na & waste_management-type_wood_waste_for_treatment_incineration for CH have been marked as "h" but there only appear to be 2 datapoints here, with rather different areas - could it not just be concidence that they have the same EF? I wonder if we are being too sensitive with this check, and should look just for where there are 3 or more across categories (2 if they are in the same category)?

This would remove h's for:

- BG fuel_type_biogasoline-fuel_use_na & other_materials-type_additives_blending_components
- AT fuel_type_motor_fuel-fuel_use_na & other_materials-type_secondary_raw_materials
- AU livestock_farming-type_pigs & metals-type_other_nonferrous_metal_products
- CN chemicals-type_chemicals_not_elsewhere_specified
- AT chemicals-type_white_spirit_sbp & fuel_type_heavy_fuel_oil-fuel_use_na

Thoughts?

maryarbab (approved):

patstirling (approved):

PR #257, opened 2022-05-24
EXIOBASE for FI FR ES GR

AnastasiaLobanova:

- [ ] New
- [x] Update

Multiple

**What?**

EXIOBASE for FI FR ES GR

**Why?**

adding all EXIOBASE


patstirling (changes requested):
Mostly clean!

following are marked h but appear to be in quite different categories?
GR: electricity-energy_source_coal & heat_and_steam-type_steam_hot_water_supply_services
GR: fuel_type_petroleum_coke-fuel_use_na & chemicals-type_bitumen
FR: fuel_type_biodiesel-fuel_use_na & other_materials-type_additives_blending_components
ES: fuel_type_gasoline_jet-fuel_use_aviation is h-linked to a lot of mined materials - this is a refined product though so may it should be considered

PR #259, opened 2022-05-25
adding EXIOBASE for PT RO RU SE SI SK TR TW WA WE WF WL WM ZA

AnastasiaLobanova:

- [x] New
- [ ] Update

Multiple

**What?**

adding EXIOBASE for PT RO RU SE SI SK TR TW WA WE WF WL WM ZA

**Why?**

LAST part of EXIOBASE ingress


patstirling (approved):
Looks good - couple of notes taken for follow up

patstirling (approved):

PR #261, opened 2022-05-30
Add schema-based validation

GeeWee:
This PR adds a new schema-based validation to the OEFDB, which will hopefully be easier to change as the database updates.

This new check uses the file in `metadata/schema.toml` to check what columns are valid. In the future I am hoping that we can primarily edit the `schema.toml` file when columns are altered.

It will currently run the new schema-based validation along with the existing validation - both must pass for the check to be considered successful.

Example of error output:
image

### DIFF
--- a/metadata/schema.toml
+++ b/metadata/schema.toml
@@ -3,6 +3,7 @@
 # It currently contains a list of columns, and what data is valid for each column
 # Lines starting with "#"'s are comments
 
+# Explicit documentation for the schema and the validators are available here: https://github.com/climatiq/python-oefdb-sdk#schema-validation
 # See the description of the first column for the structure of the column
 
 
--- a/metadata/schema.toml
+++ b/metadata/schema.toml
@@ -331,6 +331,7 @@ allowed_values = [
     "h",
     "o|h",
     "m",
+    "e",
 ]
 
 [[columns]]
--- a/metadata/schema.toml
+++ b/metadata/schema.toml
@@ -1,0 +1,125 @@
+[[columns]]
+name = "sector"
+validators = ["is_allowed_string", "is_ascii"]
+allow_empty = false
+
+[[columns]]
+name = "category"
+validators = ["is_allowed_string", "is_ascii"]
+allow_empty = false
+
+[[columns]]
+name = "activity_id"
+validators = ["is_allowed_string", "is_ascii"]
+allow_empty = false
+
+[[columns]]
+name = "name"
+validators = ["is_allowed_string", "is_ascii"]
+allow_empty = false
+
+[[columns]]
+name = "activity_unit"
+validators = ["is_allowed_string", "is_ascii"]
+allow_empty = false
+
+[[columns]]
+name = "kgCO2e-AR5"
+validators = ["is_float_or_not_supplied"]
+allow_empty = true
+
+[[columns]]
+name = "kgCO2e-AR4"
+validators = ["is_float_or_not_supplied"]
+allow_empty = true
+
+[[columns]]
+name = "kgCO2"
+validators = ["is_float_or_not_supplied"]
+allow_empty = true
+
+[[columns]]
+name = "kgCH4"
+validators = ["is_float_or_not_supplied"]
+allow_empty = true
+
+[[columns]]
+name = "kgN2O"
+validators = ["is_float_or_not_supplied"]
+allow_empty = true
+
+[[columns]]
+name = "kgCO2e-OtherGHGs-AR5"
+validators = ["is_float_or_not_supplied"]
+allow_empty = true
+
+[[columns]]
+name = "kgCO2e-OtherGHGs-AR4"
+validators = ["is_float_or_not_supplied"]
+allow_empty = true
+
+[[columns]]
+name = "uncertainty"
+validators = ["is_int"]
+allow_empty = true
+
+[[columns]]
+name = "scope"
+validators = []
+allow_empty = true
+
+[[columns]]
+name = "lca_activity"
+validators = ["is_allowed_string", "is_ascii"]
+allow_empty = false
+
+[[columns]]
+name = "source"
+validators = ["is_allowed_string", "is_ascii"]
+allow_empty = false
+
+[[columns]]
+name = "year_released"
+validators = ["is_year"]
+allow_empty = false
+
+
+[[columns]]
+name = "years_valid"
+validators = ["is_allowed_string", "is_ascii"]
+allow_empty = true
+
+[[columns]]
+name = "years_calculated_from"
+validators = ["is_year"]
+allow_empty = true
+
+[[columns]]
+name = "region"
+validators = ["is_allowed_string", "is_ascii"]
+allow_empty = false
+
+[[columns]]
+name = "data_quality"
+validators = []
+allow_empty = false
+
+[[columns]]
+name = "contributor"
+validators = ["is_allowed_string"]
+allow_empty = true
+
+[[columns]]
+name = "date_accessed"
+validators = ["is_date"]
+allow_empty = false
+
+[[columns]]
+name = "description"
+validators = ["is_allowed_string"]
+allow_empty = false
+
+[[columns]]
+name = "source_link"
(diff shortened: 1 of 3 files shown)


grindcode (dismissed):
LGTM!

metadata/schema.toml
```diff
@@ -0,0 +1,139 @@
...
+
+[[columns]]
+name = "region"
+validators = ["is_allowed_string", "is_ascii"]
+allow_empty = false
+
+[[columns]]
+name = "data_quality"
+validators = []
+allow_empty = false
```
AnastasiaLobanova, line 119: I think this column can be empty - it means the data quality is ok, it contains flags if there is something wrong

metadata/schema.toml
```diff
@@ -0,0 +1,139 @@
...
+
+[[columns]]
+name = "years_valid"
+validators = ["is_allowed_string", "is_ascii"]
+allow_empty = true
+
+[[columns]]
+name = "years_calculated_from"
+validators = ["is_year"]
```
maryarbab, line 108: shouldn't "years_calculated_from" be also a string, in case there are multiple years?

metadata/schema.toml
```diff
@@ -0,0 +1,139 @@
...
+allow_empty = true
+
+[[columns]]
+name = "scope"
+validators = []
+allow_empty = true
+
+[[columns]]
+name = "lca_activity"
+validators = ["is_allowed_string", "is_ascii"]
```
maryarbab, line 88: @AnastasiaLobanova @patstirling shouldn't we also have "not-supplied" in case the LCA activity is unknown (i.e. such methodology is not supplied by the source)?

maryarbab (changes requested):

metadata/schema.toml
AnastasiaLobanova: I think probably the broader validator of `is_allowed_string` also includes the `not-supplied` string, but Gustav can say for sure :)

metadata/schema.toml
GeeWee: I think you'er right!

metadata/schema.toml
```diff
@@ -0,0 +1,139 @@
...
+allow_empty = false
+
+[[columns]]
+name = "activity_id"
+validators = ["is_allowed_string", "is_ascii"]
+allow_empty = false
+
+[[columns]]
+name = "name"
```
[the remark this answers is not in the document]
GeeWee: Unsure - `is_allowed_string` basically just means "this is a non-empty string without a comma"

Currently there are a bunch of names (30 or so) that are non-ascii, so I just adhered to that when creating the schema. Should `name` be ASCII only?

`activity_id` needs to be ascii-only, we rely on that in the API

metadata/schema.toml
```diff
@@ -0,0 +1,139 @@
...
+validators = ["is_allowed_string"]
+allow_empty = true
+
+[[columns]]
+name = "date_accessed"
+validators = ["is_date"]
+allow_empty = false
+
+[[columns]]
+name = "description"
```
AnastasiaLobanova, line 132: Here we have sometimes funky french (sorry!) characters, e.g. in the descriptions of ADEME or UBA...would this be a problem here for this validator?

metadata/schema.toml
GeeWee: Might be! Currently it seems like we only have a single-year entries. I think the good part is that if we then _do_ add something that contains multiple years, we can just edit the schema. I'm open to doing it now if we expect that we'll get multiple years in there soon though.

metadata/schema.toml
GeeWee: Yes we do! No, `allowed_string` basically just says "no commas" - perhaps we should rename it to "no_commas"? - if we checked for ASCII in description we'd end up with a lot of issues (I tried that at one point :D)

metadata/schema.toml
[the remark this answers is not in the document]
AnastasiaLobanova: Ah ok now I get it - I thought allowed string was alphanumeric string with only underscores and hyphens! All good then! 

Then also my comment below is non valid :)

GeeWee:
Just noting something here - the image is just an example of output that failed back when I tried to validate `name` with the `is_ascii` validator. Currently as you can see with the "1 successful check", the validator validates the OEFDB as-is.

metadata/schema.toml
AnastasiaLobanova: Hehe! Yes, sorry I got confused :) All good then from my side :)

AnastasiaLobanova (dismissed):

metadata/schema.toml
GeeWee: `is-allowed-string` is basically just "any string that doesn't contain commas" - but I can see I'll need to rename it a little later to `no_commas` as it's obviously not very good at saying what it actually checks for :D

metadata/schema.toml
maryarbab: sounds good! yup, let's keep it as it is and we can change it if have multiple years.

metadata/schema.toml
maryarbab: Yes, but what I actually meant was what we currently have in the DB for these fields in case they are not provided by the source. We can discuss it in the data sync I guess

AnastasiaLobanova: Ah ok - I see! Let's discuss this in the sync indeed

GeeWee:
@maryarbab requested re-review, not sure if there's still things I need to do here from your end?

maryarbab (approved):

metadata/schema.toml
```diff
@@ -0,0 +1,350 @@
...
+
+
+# The [[columns]] denotes the start of a new column
+# The columns are ordered - it is checked that the columns are in the same order in this file and in the CSV file
+[[columns]]
+# This is the name of the column - aka the name that is in the header row
+name = "sector"
+# This is the list of "validators" to apply to each value in the column. Each validator will validate some aspect of of the value
+# All validators must pass for a value to be considered valid.
+validators = ["has_no_commas", "is_ascii"]
```
GeeWee, line 17: Notice that `is_valid_string` has changed name to `has_no_commas` and `is_ascii` is new.

metadata/schema.toml
```diff
@@ -0,0 +1,350 @@
...
+# Most validators expect that the CSV value is not-empty. If you want to allow *both* empty values, and values that pass
+# all validators, set allow_empty to be true. Then empty values will always be considered valid
+allow_empty = false
+# If there is a pre-defined set of values that are allowed in a field you can add the "allowed_values"
+# attribute. Note that allow_empty still counts, so if allow_empty is set to true, the cell does not have to be one
+# of the allowed values.
+allowed_values = [
```
GeeWee, line 24: Note that there is new documentation about how allow_empty works with validators., and the new `allowed_values` feature.

Unfortunately it was a little cumbersome to use the sector-guidance csv sheet. If we find out that having these sectors/categories in more than one place, we can revisit later.

metadata/schema.toml
```diff
@@ -0,0 +1,350 @@
...
+    "Paper and Cardboard Waste",
+    "Plastic Waste",
+    "Waste Management",
+    "Water Supply",
+    "Water Treatment",
+]
+
+[[columns]]
+name = "activity_id"
+validators = ["is_valid_activity_id", "is_ascii"]
```
GeeWee, line 148: New validator here: `is_valid_activity_id` - that checks for alphanumerics and the special id chars allowed.

GeeWee (commented):
Hey team! Can I get another pass on the new schema?

maryarbab (dismissed):

metadata/schema.toml
AnastasiaLobanova: Hehe - now since those characters really became a problem - we could delete all of them and then really try to check for ASCII in the description? If it won't be too much of a clean up

AnastasiaLobanova (approved):

AnastasiaLobanova (approved):

PR #270, opened 2022-06-21
Add room-night to allowed unit values

grindcode:

- [ ] New
- [x] Update

None

**What?**

Update validation schema

**Why?**

Validator was failing due to change of unit in hotel entries

### DIFF
--- a/metadata/schema.toml
+++ b/metadata/schema.toml
@@ -180,6 +180,7 @@ allowed_values = [
     "GB-hour",
     "GB",
     "TB-hour",
+    "person-night",
     "room-night",
     "tonne-km",
     "ton-mile",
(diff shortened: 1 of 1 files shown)


maryarbab (dismissed):

GeeWee:
Validator update looks great, but OEFDB still has one error - seems like `person-night` still exists in one column.
Suggest we fix before merging.

GeeWee (approved):

patstirling:
This isn't a bug - we need to support both `room-night` and `person-night`

GeeWee:
Ah, then it looks good now!

GeeWee (approved):

PR #288, opened 2022-08-01
adding BEIS 2022 part 1

AnastasiaLobanova:

- [x] New
- [ ] Update

Multiple

**What?**

adding BEIS 2022 part 1

**Why?**

Annual update of key data sources

### DIFF
--- a/.~lock.OpenEmissionFactorsDB.csv#
+++ b/.~lock.OpenEmissionFactorsDB.csv#
@@ -1,1 +1,0 @@
-,anastasialobanova,Anastasias-Air,01.08.2022 11:29,file:///Users/anastasialobanova/Library/Application%20Support/LibreOffice/4;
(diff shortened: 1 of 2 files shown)


patstirling (changes requested):
Can't review in diff as it's too large, so comments here:
- we aren't populating the "year_calculated" field, but we did for 2021 (although weirdly said is was calculated from 2021 which isn't possible). We should either:
1. Populate the column for 2022 data with "2021" and (in another PR) update the 2021 data with "2020" or
2. Remove data from this column in the 2021 dataset (in another PR)

- Homeworking - change unit to "hour" and qualify that it is "FTE working hour" in the description.

maryarbab (dismissed):

patstirling (dismissed):

maryarbab (approved):

PR #289, opened 2022-08-01
adding Energy sector from BEIS 2022

AnastasiaLobanova:

- [x] New
- [ ] Update

Energy Sector

**What?**

adding Energy sector from BEIS 2022

**Why?**

Annual update of key data sources


maryarbab (approved):

patstirling (commented):
I'll not request changes but here's things I spotted:
- some non-standard hyphens in the name field (typically before "WTT") - being decoded as ‚Äì in excel (whereas normal hyphens are fine)
- bunch of duplicates - do you want help identifying them?

PR #294, opened 2022-08-08
Add GLEC default road EFs and update ranges

patstirling:

- [ ] New
- [x] Update

Road Freight

**What?**

Add GLEC's default road EFs and update vehicle weight ranges to include lower bounds where there are overlapping ranges

**Why?**

To allow for defaulting to basic EFs according to the GLEC framework and avoid logic clashes when tonnage of vehicle is provided.


OpenEmissionFactorsDB.csv
```diff
@@ -11427,6 +11431,10 @@ c1225502-2fcf-4a1c-8e3e-4066096c63b6,Transport,Road Freight,freight_vehicle-vehi
...
+,Transport,Road Freight,freight_vehicle-vehicle_type_mgv_refrigerated-fuel_source_na-vehicle_weight_gt_7.5_lt_20-distance_uplift_included,MGV (7.5-20 t) - Refrig/temp controlled,tonne-km,not-supplied,#VALUE!,not-supplied,not-supplied,not-supplied,not-supplied,not-supplied,,,well_to_wheel,GLEC,2019,2019 - 2022,,ASIA_AFRICA,m,Climatiq,2022/06/20,Well-to-Tank (WTT) emission intensity of freight. Retrieved from the Global Logistics Emissions Council Framework for Logistics Emissions Accounting and Reporting. (2019). NOTE: The source does not clarify if the kgCO2e value is calculated using either IPCC Fourth Assessment Report (AR4) or IPCC Fifth Assessment Report (AR5) methodologies.This factor includes a +5% distance adjustment to allow for minor deviations and delays to the planned route.,https://www.smartfreightcentre.org/en/how-to-implement-items/what-is-glec-framework/58/
```
AnastasiaLobanova, line 11436: Here is something wrong with the EF value

AnastasiaLobanova (changes requested):

OpenEmissionFactorsDB.csv
patstirling: Ay carumba, google sheets formula fail! Ok, values should be fine now.

AnastasiaLobanova (approved):
Looks good!

PR #297, opened 2022-08-16
correction BEIS 2021 numbers

AnastasiaLobanova:

- [ ] New
- [x] Update

multiple

**What?**

correction of BEIS 2021 CH4 and N2O numbers

**Why?**

they were calculated wrongly


maryarbab (approved):
I did some random checks and they all look correct. Thanks @AnastasiaLobanova !

PR #298, opened 2022-08-23
adding new data from CCF

AnastasiaLobanova:

- [x] New
- [ ] Update

Information and Communication sector

**What?**

adding new data from CCF

**Why?**

annual update of key sources

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -3,6 +3,7 @@ package.json
 node_modules
 .DS_Store
 *.bak
+.~lock.OpenEmissionFactorsDB.csv#
 
 # Generated in CI
 python-oefdb-sdk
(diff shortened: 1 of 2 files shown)


maryarbab (changes requested):
Great work @AnastasiaLobanova! A few notes:

1. I'd suggest to remove all years from the year columns and just keep the data_accessed as a reference to the date. The source is a website and we can not track how often they update their assumptions, so we do not have an accurate estimates of year_released, year_valid, and year_calculated. We need to do this for both CCF 2021 and CCF 2022 and update the [source table](https://github.com/climatiq/Open-Emission-Factors-DB/blob/main/OEFDB_sources.md). 
2.  I'd suggest to add a "m" flag to all CCF EFs and add this note to the description: The source does not clarify if the kgCO2e value is calculated using either IPCC Fourth Assessment Report (AR4) or IPCC Fifth Assessment Report (AR5) methodologies.
3. AZURE and AWS numbers seem to be different than the ones calculated in the Gsheet. Let's check those.

PR #301, opened 2022-08-26
fuel EFs from GEMIS

AnastasiaLobanova:

- [x] New
- [ ] Update

Energy sector

**What?**

fuel EFs from GEMIS

**Why?**

missed EFs from the last GEMIS ingress


maryarbab (approved):
Looks clean!

PR #305, opened 2022-09-06
year_valid column

maryarbab:

- [*] Update

_Multiple_

**What?**

_Populate the year_valid column for all sources_

**Why?**

_Update the OEFDB with the valid year for each EF._


AnastasiaLobanova (approved):
Looks good Maryam! GREAT effort!

PR #306, opened 2022-09-16
Format and order OEFDB

GeeWee:

- [ ] New
- [x] Update

All

**What?**

This PR runs the new OEFDB formatter that automatically both ensures a consistent formatting and ordering of the OEFDB before merging in a pull request 

The ordering is a little different than the current one, but I think it's equally valid, and it should be consistent.

 

This is just a PR to run the new automatic ordering script in the oefdb-sdk


GeeWee:
Hey team. Notice that the formatting comment has changed as well, to contain both "schema" fixes such as wrong value, and formatting changes which are ordering and formatting-dependant. Schema changes cannot always be fixed, but formatting can always be done automatically.

Seeing as this reorders some things, we might have to coordinate with any PR's that are still hanging around based on the old sort, but I'm sure we can figure out a solution for that.

Please do review that the sorting makes sense, and that it hasn't deleted any lines or similar. It shouldn't have, but some extra eyes on this is fine.

I don't expect everyone who I've tagged as a reviewer to thoroughly review this, but I would like everyone to know what's going on.

AnastasiaLobanova (dismissed):
I think the sorting looks good.

This work is of GREAT help to Science and data team! Thank you Gustav!!

GeeWee:
You're welcome! I've just pulled in the latest main so it should be merge-able again

AnastasiaLobanova (approved):
Approved

maryarbab (approved):
A big help to the data team! Thank you, @GeeWee!

PR #309, opened 2022-09-21
EEA data ingress

maryarbab:

- [*] New

_Electricity_

**What?**

_Add European grid mix data_

**Why?**

_To provide a better coverage of European grid mix EFs_


AnastasiaLobanova (approved):
Looks good Maryam! Approved

PR #324, opened 2022-10-26
Run autoformatter

GeeWee:
This PR is the first one to run the new oefdb-sdk based autoformatter for number values, so that all numbers are formatted identical no matter who creates the csv.

Many rows should change representation - but the value of no row should change.


GeeWee:
@AnastasiaLobanova / @maryarbab  - please spot check the diff to see if there are any inconsistencies. Spot checks from my side looks good.

AnastasiaLobanova (approved):
Looks good!

maryarbab (approved):

PR #330, opened 2022-11-02
Year_valid updates

maryarbab:

- [*] Update

_Multiple_

**What?**

_Correct year_valid for GHG protocol, UNFCCC, and Refrigerant gases_

**Why?**

_To reflect years that the above datasets are valid for_


GeeWee:
Is this PR ready for me to diff it regarding year_valid_from vs year_published?

maryarbab:
@GeeWee yes, it is ready!

AnastasiaLobanova (changes requested):
Thank you Maryam! I have some comments:

1. For refrigerants GHG Protocol: shouldn't be year_calculated_From be simply 2007? Instead of 2007-2022. They are all from AR4? If also from AR5, then probably a range of 2007-2013 would work better. 
2. Line 12388 - this entry has year_calculated ranging from 2013-2022 - is this value from AR5? Should we simply put 2013?

UNFCC and GHG Protocol - look good to me

Another general comment: there are ~6k changes in the first commit to PR, mainly bc gsheets have changed the format of scientific notation as well as turned the representation of 0 from double to an integer. Then formater changed these values back but it is impossible to separate changes made by him from changes to review in github, which made it hard to review the changes.
I have made some tests - if we untick the box of "converting test to numbers, dates and formulas" at the import of .csv to Gsheets, then he treats our numbers as text and doesn't introduce changes to the scientific format and doesn't turn doubles to integers.

GeeWee:
@AnastasiaLobanova just out of curiousity - do you review each commit separately?

GeeWee:
Just checked over the PR - did we not agree that we would make `year_valid_from` a single number and not a range?

AnastasiaLobanova:
> @AnastasiaLobanova just out of curiousity - do you review each commit separately?

Yes I am looking at commit 3ca6aa7

GeeWee:
I see. Normally I just view the entire diff between `main` and the branch - in which case it doesn't matter if the change is in one or two commits.

I've checked out the changes and made a diff currently:
https://github.com/climatiq/edge-api/pull/362

**Also keep in mind that it looks like this PR has garbled a bunch of the formatting. Lots of `�` characters.

AnastasiaLobanova:
Ok - I will take a look at all commits. Still, I think treating numbers as strings is not a bad idea to be on the safe side. Also it is just a tick to take out at the import.

Re ??? - I can probably fix them in the current PR. We have to fish these special characters out of the DB.

Re ranges - yes we did! It is just spend-based data left with some ranges, I wanted to tackle it today. It is a problem for your checks, right?

PR #331, opened 2022-11-04
Add BOM to database so that applications opening it can tell it's charset

platy:
**What?**

Trying to add a [BOM](https://en.wikipedia.org/wiki/Byte_order_mark) to the oefdb so that tools used to read it will understand it's using UTF8 and users won't see any encoding issues.

**Why?**

Currently there is a risk that applications used to read the oefdb will guess at the wrong character set encoding and some symbols and diacriticts will show incorrectly.


AnastasiaLobanova:
@platy do we want still to investigate formatting this in this PR?

platy:
No I'll close this - thanks for bringing it up.

PR #332, opened 2022-11-07
year_valid_from updates

maryarbab:

- [*] Update

_All_

**What?**

_Change year_valid tp year_valid_from across the OEFDB and update it across the wholedatase_

**Why?**

_To make sure we have year_valid_from correct across the whole DB as we are changing the year logic in the API from year_released to year_valid_from_


AnastasiaLobanova (approved):
Looks clean! Approved!

PR #335, opened 2022-11-07
moving some BEIS values from AR4 to AR5

AnastasiaLobanova:

- [ ] New
- [x] Update

multiple

**What?**

moving some BEIS values from AR4 to AR5

**Why?**

Because some data gathered to calculate the conversion factors is based on the IPCC Fifth Assessment Report (AR5) GWPs already and cannot be disaggregated into constituent gases to be converted to AR4 GWPs. 
This includes the "Hotel stay" factors and the "Material Use" and "Waste Disposal" factors for glass, electricals, batteries, asbestos, paper and board, metals, steel, aluminium, mixed cans, scrap metal and mixed waste. Those values are currently supplied to the OEFDB as AR4, this PR brings them to AR5 column


maryarbab (dismissed):
Looks clean!

maryarbab (approved):

PR #336, opened 2022-11-08
EPA spend-based data update

AnastasiaLobanova:

- [x] New
- [ ] Update

multiple

**What?**

new, updated numbers for the EPA spend-based data

**Why?**

EPA has issued a new version of the same dataset


maryarbab (changes requested):
Looks great! Just a quick note about the dataset field: Are we planning to have a specific format for it? "SupplyChainEmissionFactorsforUSIndustriesCommodities" is not very intuitive. Also, why are there many different versions for 2022 and none for 2020? Is that a typo?

maryarbab (approved):

PR #351, opened 2022-11-28
Facility properties and emissions added to the OEFDB from GHG Protoco…

deliceli:
…l Evulation Tool Scope3

New

 Facility 

**What?**

Emission factors for the "Facility" category added from GHG Evaluation Tool Scope 3. 

**Why?**

Client request


maryarbab (changes requested):
Looks good @deliceli, thank you! Few notes:

1. Does the source mention LCA boundary? If not, we should have "unknown" in the lca_activity field.
2. "Facility" should be added to the schema and also "Sector-Category-ID_structure.csv" file. You can also make those changes in this PR.
3. I wonder if we should provide more information about what each scope includes in terms of activities. Does the source provide such information?

OpenEmissionFactorsDB.csv
```diff
@@ -19079,3 +19079,45 @@ a2db0d60-2494-43d7-ac89-0c71a42a677d,Energy,Electricity,electricity-energy_sourc
...
 51711eba-7a91-4f13-9f0a-08fe7e49f0df,Energy,Electricity,electricity-energy_source_grid_mix,Grid mix,kWh,not-supplied,not-supplied,0.1,1.7e-05,2e-06,not-supplied,not-supplied,,2,electricity_generation,UNFCCC,,2021,2021,2019,CA-YT,,Climatiq,2021/07/30,Emission intensity of grid mix as reported. Published in Canada 2021 National Inventory Report (NIR). Includes generation factor.,https://unfccc.int/documents/271493
+,Buildings and Infrastructure,Facility,facility-type_lodging_owned,Lodging facility owned,m2,not-supplied,40.145,not-supplied,not-supplied,not-supplied,not-supplied,not-supplied,,3,,GHG Protocol,,2017,2016,2016,GLOBAL,,Climatiq,2022/11/25,Emission intensity of lodging facility owned. Factors retrieved from the Quantis+GHG Protocol Scope 3 Calculator for the year 2016 and apply for years beyond 2016.,https://quantis-suite.com/Scope-3-Evaluator/
```
AnastasiaLobanova, line 19082: There are two other EFs for Scopes 1 and 2 missing in the Lodging.

I also don't see Education facilities. Could you cross-check that all types of facilities are included here please

AnastasiaLobanova (changes requested):

OpenEmissionFactorsDB.csv
deliceli: All the comments taken into consideration

maryarbab (approved):

PR #353, opened 2022-12-02
Add diffdaff Visual diff builder

platy:
**What?**

Adds a github action which activates on every commit added to a PR, building a visual diff using the [daff tool](https://paulfitz.github.io/daff/) and publishing it using github pages.

A link to the diff page is added as a comment to the PR once it is ready.

Hopefully we can later build on this and make it more useful.

**Why?**

This should make it easier to review certain common changes to the OEFDB which are difficult to review in regular diffs


.github/workflows/diffdaff.yaml
```diff
@@ -0,0 +1,55 @@
...
+      group: preview-${{ github.ref }}
+      cancel-in-progress: true
+    strategy:
+      matrix:
+        python: [ 3.9.12 ]
+    steps:
+      - name: Checkout this
+        uses: actions/checkout@v2
+        with:
+          path: this
```
GeeWee, line 20: What does this do?

GeeWee: Ooh, it checks out the current PR. Can we rename action to "checkout current PR" or something? I went looking for whether or not "this" was some sort of specific argument. Only a nit, so feel free to disregard.

GeeWee (commented):
Small nit, but otherwise let's get this in!

.github/workflows/diffdaff.yaml
```diff
@@ -0,0 +1,55 @@
...
+          path: main
+      - name: Set up Python ${{ matrix.python }}
+        uses: actions/setup-python@v2
+        with:
+          python-version: ${{ matrix.python }}
+      - uses: BSFishy/pip-action@v1
+        with:
+          packages: |
+            daff
+      - run: mkdir diff && daff --unordered --output diff/temp.html main/OpenEmissionFactorsDB.csv this/OpenEmissionFactorsDB.csv
```
GeeWee, line 34: Wonder if we should try to always show UUID & activity ID, but I'm not sure. Maybe someone from the data team should weigh in.

platy: I've marked that as an item on the follow up task - I think data will have preferences once they can use it. UUID seems to always be shown based on my testing so far. I think daff uses the first column as the primary key if none is selected.
https://linear.app/climatiq/issue/PROD-222/follow-up-on-diffdaff

platy:
@maryarbab @AnastasiaLobanova @patstirling are you happy to add this feature to oefdb PRs?

maryarbab:
> 

Sounds great to me! Thanks @platy !

AnastasiaLobanova:
Will be def VERY helpful!

.github/workflows/diffdaff.yaml
GeeWee: Makes sense.

GeeWee (approved):

PR #354, opened 2022-12-02
CCF Embodied EFs - AWS

maryarbab:

- [ ] New

_Cloud Computing - CPU_

**What?**

_Add cloud computing embodied emissions_

**Why?**

_To enable full life cycle analysis of cloud computing carbon footprint._


AnastasiaLobanova (approved):
Just one suggestion and one question that I have here.

PR #357, opened 2022-12-05
Ghg evaluation

deliceli:

New

Facility

**What?**

GHG Evaluation Tool Scope 3 Facility Emission added.

**Why?**

Client request.


AnastasiaLobanova (approved):
Looks good to me! Thank you @deliceli

maryarbab (approved):
Looks clean! Thank you @deliceli !

PR #362, opened 2022-12-09
GCP & AZURE embodied EFs + updates to AWS

maryarbab:

- [X] New
- [X] Update

_Cloud Computing - CPU_

**What?**

_Add GCP & AZURE embodied emission factors and update AWS embodied EFs name field _

**Why?**

_To enable full life cycle analysis of cloud computing footprint_


AnastasiaLobanova (changes requested):
Are names for AWS correct in this PR?

AnastasiaLobanova (approved):

PR #363, opened 2022-12-14
add EFs for CT 2019, 2020, 2021 & 2022

gpantelidou:

- [ ] New
- [X] Update

Multiple 

**What?**

Added EFs from CT datasets for years 2020, 2021, 2022. Also for 2019 since some countries were missing from the OEFDB. I also had to change the schema file and include the "p|m" data quality indicator. 

**Why?**

client request


OpenEmissionFactorsDB.csv
```diff
@@ -293,204 +293,204 @@ e67fe9ef-c496-4db1-a004-590ca57ea4da,Transport,Vehicles,passenger_vehicle-vehicl
...
+0f55ed7c-6592-4dc1-8553-dca019defc3f,Energy,Electricity,electricity-energy_source_total_supplier_mix,Total supplier mix,kWh,not-supplied,0.79462,not-supplied,not-supplied,not-supplied,not-supplied,not-supplied,,,electricity_generation,AIB,,2022,2022,2021,PL,p,Climatiq,2022/10/01,Emission intensity of total supplier mix as reported. Published by the Association of Issuing Bodies (AIB). Retrieved from 2021 European Residual Mix Report. NOTE: this value is provided in CO2 only and does not include the impact of other GHGs. The source does not clarify if the kgCO2e value is calculated using either IPCC Fourth Assessment Report (AR4) or IPCC Fifth Assessment Report (AR5) methodologies.,https://www.aib-net.org/facts/european-residual-mix/2021
```
AnastasiaLobanova, line 480: If the source provides CO2 values only, why do we need a phrase of AR4 and AR5?

AnastasiaLobanova (changes requested):

OpenEmissionFactorsDB.csv
maryarbab: @AnastasiaLobanova That is true. It was my bad recommendation provided to @gpantelidou based on 2019 data. Agree to remove it. Also let's discuss in the data sync which column we should provide CO2 only values (CO2e, CO2, or both). It seems that we are not very consistent across the OEFDB.

maryarbab (dismissed):
@gpantelidou Looks great! Please just remove the note Anastasia mentioned and it is ready to go:)

maryarbab (approved):

PR #366, opened 2022-12-14
UNFCCC data Canada

AnastasiaLobanova:

- [x] New
- [ ] Update

Fuel, Electricity

**What?**

Adding Fuel and Electricity EFs

**Why?**

Complete historical series for natural gas and gridmix from NIR


maryarbab (changes requested):
Looks great @AnastasiaLobanova ! Just a couple of notes:
1- shouldn't the name also reflect the use of natural gas (i.e. utilities, industrial, ...)?
2- Are these EFs all from 2022 NIR? I see this note in Linear:
"Note that the year_calculated and year_valid_from for each year is 2 years prior to the year_released - so for example the last year released (2022) should have year_released=2022, year_valid_from=2020, and year_calculated=2020."

maryarbab (approved):

PR #372, opened 2022-12-19
fix small GLEC issues

maryarbab:

- [X] Update

_Multiple_

**What?**

_Fix some activity_ids_

**Why?**

_To remove any bugs from the GLEC dataset_


AnastasiaLobanova (approved):
Looks good!

PR #380, opened 2022-12-21
increment year released for UNFCCC natural gas +2

AnastasiaLobanova:

- [ ] New
- [x] Update

Fuel

**What?**

increment year released for UNFCCC natural gas +2

**Why?**

because of year released vs year valid unresolved conflict


patstirling (approved):
Looks clean!

PR #387, opened 2023-01-16
Fix a typo

maryarbab:

- [X] Update

_Cloud Computing - CPU_

**What?**

_Change `accelorator` to `accelerator`._

**Why?**

_Fix the typo_


GeeWee:
I mean this is the right change - it is a breaking change though. Unsure if anyone uses these or we'll just need to add it to the id deprecation list?

GeeWee:
Yeah no this looks good. Can you please link this PR to the Linear issue so we can trace it later [instructions here](https://linear.app/docs/github#link-prs)

Also - do we know why CI isn't running on this branch? @platy

GeeWee (approved):

PR #390, opened 2023-01-18
EFs for gridmix and fuels for DE

AnastasiaLobanova:

- [x] New
- [] Update

Energy sector

**What?**

adding EFs for upstream and fuel-combustion of fuels, gridmix, heat and steam

**Why?**

customer request


maryarbab (changes requested):
Looks great, thanks @AnastasiaLobanova! A few notes:

The unit for heat and electricity EFs should be kWh.
Let's remove "upstream" and "t&d" from names (I had to mention it as it also applies to the heat category not just electricity).
Let's have "electricity_generation" as the lca_activity and scope 2 for UBA grid mix data. 
The "gridmix" in descriptions needs to be changed to "grid mix".

TimCroker (commented):
Just one question - all the district heating is fossil-fuel mix. Is that definitely what they want rather than the average mix?

AnastasiaLobanova:
@maryarbab -thank you! not sure why I put scope 1 for electricity 😵‍💫  good catch with the units also. 


@TimCroker  That's what they have asked to ingress from the respective dataset (see gsheet) (and that's the only EF we have found from UBA for district heating). You can maybe ask Annika, Tim? I understand you will get in touch with her per email today

maryarbab (approved):

maryarbab (approved):

PR #391, opened 2023-01-18
add dataset names

gpantelidou:

- [ ] New
- [x] Update

dataset, description and source link columns

**What?**

added dataset names, fixed some wrong source links and added description note for archived BEIS 2018 spend-based data

**Why?**

more clear for the user


AnastasiaLobanova (dismissed):
Not sure if eGRID 2019 and eGRID 2020 should be just eGRID, following our discussion on Monday?
Otherwise looks good I think!

patstirling (changes requested):
Suggestions round one:

1. `Greenhouse gas reporting: conversion factors` -> `UK Government GHG Conversion Factors for Company Reporting` (as per the title provided within the spreadsheet)
2. `UK and England's carbon footprint to 2019` -> `UK's carbon footprint (SIC spend-based multipliers)` (removing reference to year as of course when they release it next year it will say "to 2020" and adding the pertinent information that it is spend-based data based on COI categorisation. Removing England as the actual dataset used only refers to UK)
3. `UK BEIS spend-based data (Archived`) -> `UK's carbon footprint (COICOP spend-based multipliers - archived)` (to align with above except for the specific aspects in which it differs - and that it is no longer available)

maryarbab (changes requested):
Looks great!! Thank you @gpantelidou . I have the following suggestions/comments:

- Add "retrieved from Base Carbon v20.1" to ADEME EFs descriptions
- Be consistent with the capitalization. Is it only the first word capitalized?
- Remove year from eGRID datasets.
- Remove 3 from EXIOBASE
- "Embodied emissions" is only correct for the upstream_end_of_life EFs not the use-phase ones. Suggest we use "Cloud Carbon Footprint" for all CCF EFs. 

@patstirling @AnastasiaLobanova @TimCroker let me know if you have any thoughts.

patstirling:
4. `Embodied Emissions` -> `Embodied Emissions constants` to reflect the name of the dataset as provided by CCF
5. `DEWA Sustainability Report` -> `Sustainability Report` - remove "DEWA" as this is in the source field (consistent with other areas where we have not repeated the source in the dataset name)
6.  `Annual European Union greenhouse gas inventory` -> `Greenhouse gas emission intensity of electricity generation` - I can't see a reference to the `Annual European Union greenhouse gas inventory` at the source URL? Suggest we use the term provided for the specific table that we have retrieved these data from
7. `eGrid2019` and `eGrid2020` -> `eGrid` - remove references to year - the dataset is eGrid and they release each year.
8.  `Supply Chain Factors Dataset` -> `Supply Chain Factors Dataset (commodities - spend-based)` to ensure it is clear that the specific dataset was commodities not industries, and to provide the user with context regarding spend-based (per UK BEIS example)
9. It appears the the appropriate source for GEMIS is IINAS? We can't fix this now but suggest we do so at next version of the API @AnastasiaLobanova @maryarbab ?
10. `GLEC Framework` -> `Default Fuel Efficiency and CO2e Intensity Factors` as this is the name of the module that presents the emission factors, and accurately describes this set of EFs

AnastasiaLobanova:
@patstirling good call for GEMIS -INAS, was also thinking about it! Will add to triage
@maryarbab agree to use Cloud Carbon Footprint" for all CCF EFs.

patstirling:
> 

cool - another thing that this highlights with source names is that we sometimes put the UN-LOCODE of a country in front of their ministry (e.g. UK BEIS) and sometimes we don't (EPA, MfE, DISER). This despite the EPA, for example, calling itself "US EPA". So I think it would make it nice and clear if for governmental bodies we add the code. Another one for the next version of the API?

patstirling:
> "Embodied emissions" is only correct for the upstream_end_of_life EFs not the use-phase ones. Suggest we use "Cloud Carbon Footprint" for all CCF EFs.

Good point @maryarbab - although we could provide the underlying datasets that each of the various types came from? This highlights that the embodied ones come directly from a sheet of EFs (I think??) whereas the others are one that we've factored ourselves. We should probably have "Derived from CCF methodology using XXX" for the Compute/Memory/Network/Storage EFs where XXX is the set of EFs that are provided?

patstirling:
> Be consistent with the capitalization. Is it only the first word capitalized?

I actually think we should reflect whatever the source does here @maryarbab - although I know we normalise fields like "name" elsewhere so I can be convinced :)

maryarbab:
> > "Embodied emissions" is only correct for the upstream_end_of_life EFs not the use-phase ones. Suggest we use "Cloud Carbon Footprint" for all CCF EFs.
> 
> Good point @maryarbab - although we could provide the underlying datasets that each of the various types came from? This highlights that the embodied ones come directly from a sheet of EFs (I think??) whereas the others are one that we've factored ourselves. We should probably have "Derived from CCF methodology using XXX" for the Compute/Memory/Network/Storage EFs where XXX is the set of EFs that are provided?

@patstirling We developed calculations to derive both sets of EFs (use-phase and embodied emissions).

patstirling:
Based on discussion:

- capitalisation should stay as per source

UK spend-based:

`UK full dataset including conversion factors by SIC code`
`UK full dataset including conversion factors by COICOP code (archived)`

EXIOBASE:

Use `EXIOBASE 3` and add "Version 3.8.2" to description field

CCF:

`Derived from CCF models`

EPA:

`Supply Chain Factors Dataset (commodities)` - add version to description (2020 no version, 2021 should be v1.1)

maryarbab (approved):

patstirling (approved):

PR #393, opened 2023-01-19
Update electricity EFs names

maryarbab:

- [X] Update

_Electricity_

**What?**

_Update EFs names_

**Why?**

_To make the naming of electricity EFs consistent across the OEFDB_


TimCroker (commented):
@maryarbab 
Most of the electricity factors by default exclude T&D so I think the GEMIS ones (e.g. below) should be renamed to simply "Electricity supplied from grid" to maintain consistency. The ones that **_do_** include T&D could be renamed "Electricity supplied from grid including T&D losses"

**Example of EF excluding T&D**
e7774a91-5170-4b63-acb7-e864c3839016 | Energy | ... | electricity-energy_source_grid_mix | Grid mix (excluding T&D)→Electricity supplied from grid (excluding T&D)

**Example of EF including T&D**
16c73621-321d-4947-9a03-751a85c0623d | Energy | ... | electricity-energy_source_grid_mix | Grid mix→Electricity supplied from grid

OpenEmissionFactorsDB.csv
```diff
@@ -1130,76 +1130,76 @@ c46bae34-9eea-451a-a1a1-a15596ba9332,Energy,Electricity,electricity-energy_sourc
...
+9917acf4-56db-4329-8aad-fc30f9b8e40b,Energy,Electricity,electricity-energy_source_grid_mix_losses,Electricity supplied from grid plus T&D losses,kWh,not-supplied,0.01769,0.0175,2.8e-06,4.027e-07,not-supplied,not-supplied,,3,transmission_and_distribution,BEIS,UK Government GHG Conversion Factors for Company Reporting,2022,2022,,GB,,Climatiq,2022/07/07,Emission intensity of transmission & distribution of UK electricity as reported. These factors include only direct emissions at UK power stations and from autogenerators plus those from the proportion of imported electricity. They do not include emissions resulting from production and delivery of fuel to these power stations. Retrieved from the Conversion Factors 2022: flat file published by the UK BEIS/Defra at the source URL.,https://www.gov.uk/government/publications/greenhouse-gas-reporting-conversion-factors-2022
```
AnastasiaLobanova, line 1133: these are just losses right? I think "Electricity supplied from grid plus T&D losses" is misleading here, sounds like it is an ef for a grid mix, including losses.
What about "Electricity supplied from grid: T&D losses "?

AnastasiaLobanova: Applies also to entries below

AnastasiaLobanova (changes requested):

OpenEmissionFactorsDB.csv
maryarbab: good point! Also, same with the UBA ones, it is misleading as those are all just about upstream and t&d

OpenEmissionFactorsDB.csv
```diff
@@ -21581,31 +21581,31 @@ ba8f9f80-042f-4096-a6df-ee7a6d39463d,Energy,Fuel,fuel_type_petrol_biofuel_blend-
...
+76aa6f8a-b3a5-4fd8-8d1b-d529ec3f5cf0,Energy,Electricity,electricity-energy_source_grid_mix,Electricity supplied from grid: T&D losses,kWh,not-supplied,0.056738,0.032599,0.000808,1.3e-05,not-supplied,not-supplied,,3,upstream-transmission_and_distribution,UBA,Emissionsbilanz erneuerbarer Energietraeger,2017,2017,,DE,,Climatiq,2023/01/17,Emission intensity of upstream fuel production and transmission & distribution of electricity grid mix as reported by UBA. Retrieved from Emissionsbilanz erneuerbarer Energieträger Report 2017.,"https://www.umweltbundesamt.de/publikationen/emissionsbilanz-erneuerbarer-energietraeger-2017#:~:text=Im%20Ergebnis%20weist%20die%20Netto,Stromsektor%20entfielen%20135%2C2%20Mio."
```
AnastasiaLobanova, line 21608: should we add here T&D losses plus WTT for the case of those that are upstream-t&d?

AnastasiaLobanova: Otherwise it sounds it is upstream for T&D? :D

AnastasiaLobanova (approved):

PR #397, opened 2023-01-23
CFF updates+Facility EFs updates

maryarbab:

- [X] Update

_Multiple_

**What?**

_Changed CCF embodied emissions units/updated CCF use-phase emissions descriptions and facility EFs lca_activity_

**Why?**

_Tackle small issued for the Jan release_


AnastasiaLobanova (approved):
Looks good!

PR #400, opened 2023-01-24
CCF GCP embodied emissions updates

maryarbab:

- [X] Update

_Cloud Computing - CPU_

**What?**

_Update GCP embodied emissions activity_ids_

**Why?**

_To follow the correct structure of activity_ids_


GeeWee:
Yes that looks good to me!

GeeWee (approved):

PR #406, opened 2023-01-31
add EEA 2021 data and small fixes on previous years

gpantelidou:

- [ ] New
- [x] Update

Multiple

**What?**

added new electricity data from EEA for year 2021 but also fixed some EF values from previous years

**Why?**

added new data and updated previous ones to be consistent with the source


maryarbab (approved):
Looks neat!

AnastasiaLobanova (approved):
Good job Georgia!

PR #409, opened 2023-02-03
Data flag changes

deliceli:

Update

Electricity

**What?**

Data cells for AT changed from p to e

**Why?**

To define more reasonable data to clients


AnastasiaLobanova (commented):
Looks good!

OpenEmissionFactorsDB.csv
```diff
@@ -484,11 +484,11 @@ a87a8413-f75b-46b4-931b-8dc697920aee,Energy,Electricity,electricity-energy_sourc
...
+7ebdf5f2-8616-4d8d-b084-bb8b7a62d30e,Energy,Electricity,electricity-energy_source_residual_mix,Electricity supplied from grid - residual mix,kWh,not-supplied,0.0,not-supplied,not-supplied,not-supplied,not-supplied,not-supplied,,2,electricity_generation,AIB,European Residual Mix,2022,2021,2021,AT,e,Climatiq,2022/10/01,Emission intensity of residual mix as reported. Published by the Association of Issuing Bodies (AIB). Retrieved from 2021 European Residual Mix Report. NOTE: this value is provided in CO2 only and does not include the impact of other GHGs. Note that the AIB does publish a residual factor of zero but this should not be used for calculations as it would give incorrect results. The Total Supplier Mix factor is the next best available option in the GHG Protocol market-based hierarchy of factors.,https://www.aib-net.org/facts/european-residual-mix/2021
```
maryarbab, line 487: I would have it this way just to be consistent with the previous sentence: "Note: the AIB does publish a residual factor of zero but this should not be used for calculations as it would give incorrect results. The Total Supplier Mix factor is the next best available option in the GHG Protocol market-based hierarchy of factors/"

maryarbab (changes requested):

OpenEmissionFactorsDB.csv
deliceli: Should i also change it to the capital letter to make it more consistent?

maryarbab: yes, please!

maryarbab (approved):

PR #411, opened 2023-02-03
Updates to GLEC EFs descriptions

maryarbab:

- [X] Update

_Transport_

**What?**

_Updates to descriptions_

**Why?**

_Requested by GLEC for the accreditation process_


patstirling (approved):
Looks good to me!

PR #412, opened 2023-02-06
add missing EFs for electricity aware computing endpoint

gpantelidou:

- [x] New
- [x] Update

OEFDB and schema files

**What?**

added missing data for the electricity-aware computing endpoint project

**Why?**
we need the data for accurate calculations of CPU electricity use


OpenEmissionFactorsDB.csv
```diff
@@ -21651,3 +21651,61 @@ dc0ba7ab-9069-4293-8818-7ed78e3675a0,Energy,Fuel,fuel_type_natural_gas-fuel_use_
...
+8f5c3180-39bc-4ceb-afdd-202e898d15d8,Energy,Electricity,electricity-energy_source_grid_mix,Electricity supplied from grid,kWh,not-supplied,0.76,not-supplied,not-supplied,not-supplied,not-supplied,not-supplied,,2,electricity_consumption,DISER,National Greenhouse Account Factors,2021,2021,2019,AU,,Climatiq,2023/02/06,Emission intensity of grid mix as reported by the Australian Government 2021 report.,https://www.dcceew.gov.au/climate-change/publications/national-greenhouse-accounts-factors-2021
```
patstirling, line 21657: Let's add the most recent, and complete sets of AU EFs here - it's also not a huge effort to cover the Australian state EFs. It would look a bit strange to have just one EF for Australia one year, but a full set for another year. Source URL should be: https://www.dcceew.gov.au/climate-change/publications/national-greenhouse-accounts-factors

Suggest we add all 2021 and 2022. Note that years_calculated_from will be the year of publishing and the year before, as the reporting year ends in June (e.g. for year_released=2022, years_calculated_from=2021|2022)

Note that the source has changed it's name - this is a breaking change so please keep as DISER for all years until we have an approach to data versioning which will solve for this.

I know this is a little more effort, but it's worth updating these EFs for both this endpoint and for the other uses of this data while we are looking at these reports.

GeeWee:
Just imported and there might be two EF's that are identical. I get the error
```
[ID] Duplicate composite ids found: 'electricity-energy_source_grid_mix-2022-DISER-ElectricityConsumption-KWh-AU-WA'
```
Which seems to suggest we have  two `electricity-energy_source_grid_mix` for 2022 from DISER for `AU-WA`

gpantelidou:
@GeeWee ah yes you are right I will fix that. I didn't notice that for the same source on previous year you used different activity_ids to avoid duplicates. They will be: 
`electricity-energy_source_swis_grid_mix` 
`electricity-energy_source_nwis_grid_mix`

patstirling (approved):

GeeWee:
> @GeeWee ah yes you are right I will fix that. I didn't notice that for the same source on previous year you used different activity_ids to avoid duplicates. They will be: `electricity-energy_source_swis_grid_mix` `electricity-energy_source_nwis_grid_mix`

They should be deduplicated on years - e.g. it should be possible to have different EF's with different `year_released` and the same activity_id. Perhaps @AnastasiaLobanova can elaborate more on what's going on - I don't have a good grasp of these particular EF's :)

GeeWee:
Okay, so I've finally gotten the EF selection logic working here with this branch. I don't think we all these factors, but I've found no mistakes in the activity_ids this time 'round. Happy for this to be merged into the OEFDB!

PR #413, opened 2023-02-09
delete m flag from GLEC EFs

AnastasiaLobanova:

- [ ] New
- [x] Update

Transport-related

**What?**

delete m flag from GLEC EFs

**Why?**

we know GLEC used AR4 methodology


patstirling:
Solid

patstirling (approved):

PR #414, opened 2023-02-10
adding s-flag to Shiseido data

AnastasiaLobanova:

- [ ] New
- [x] Update

Personal care

**What?**

adding s-flag to Shiseido data

**Why?**

to indicate it is a self-reported data


patstirling (approved):
Golden

PR #418, opened 2023-02-15
add electricity EFs for endpoint computing and small fixes

gpantelidou:

- [x] New
- [x] Update

Electricity

**What?**

add missing EFs for electricity endpoint computing, small fixes, also EFs for HK 2019

**Why?**

useful EFs for the endpoint computing project and now new data in the OEFDB


AnastasiaLobanova (changes requested):
Looks great Georgia! Just one thing - could you please add "not-supplied" to DISER/HK data for constituent gases and CO2e for AR5? to make it consistent with the DB?

I know we have parts of the DB with empty cells, this is something that we wanted to tackle retrospectively, but had no time to do yet. But would be good to be consistent when we are ingressing new data. Ok?

AnastasiaLobanova (approved):
Looks great!

PR #419, opened 2023-02-20
add m flag in GHGP data

gpantelidou:

- [ ] New
- [x] Update

All the categories that GHG Protocol covers

**What?**

Flag the data with a methodological `m` warning and add appropriate descirption note. Please remember that some GHGP data already have other flags like `e` or `p.` 

**Why?**

GHG Protocol removed the data from their website - so we try to notify the users


AnastasiaLobanova (approved):
looks good to me!

patstirling:
Ah - these flags should no apply to the GHG/quantis dataset (they are still hosted). Also could we please change "WARNING" to "NOTE" and change "replaced" to "removed"?

patstirling (changes requested):
See comment

gpantelidou:
Oh! I thought it was for all the data that we cite under GHG protocol. Thanks @patstirling I will fix this now.

patstirling (approved):

PR #425, opened 2023-03-10
added BEIS Fuel EF data for natural gas gross for years 2016-2020

gpantelidou:

- [x] New
- [x] Update

Fuel

**What?**

added BEIS Fuel EF data for natural gas gross for years 2016-2020

**Why?**

client request


maryarbab (approved):
Looks clean!

PR #427, opened 2023-03-27
fixed multiple errors on activity_ids

gpantelidou:

- [ ] New
- [x] Update

multiple

**What?**

fixed multiple errors on activity_ids

**Why?**

correct breaking changes


OpenEmissionFactorsDB.csv
```diff
@@ -9614,6 +9614,10 @@ b2e3cf7c-a907-464c-8971-c6fbb95b2efa,Insurance and Financial Services,Insurance
...
+d2b3c2b4-aab4-4c3c-a634-03701ac80d69,Materials and Manufacturing,Building Materials,cement_and_concrete-concrete_pipe_bricks_blocks,Concrete pipe/bricks and blocks,USD,not-supplied,not-supplied,0.461,0.002,0.0,not-supplied,0.004,,3,cradle_to_shelf,EPA,Supply Chain Factors Dataset (commodities),2022,2018,2012,US,,Climatiq,2022/10/25,Emission intensity of supply chain (with margins i.e. cradle to shelf) in US dollars spend on: concrete pipe/bricks and blocks. This factor is representative of the described commodity (equivalent to a goods or services category) and was obtained from the USEEIO model v1.0.1. Refer to the source for the source-specific data quality assessment. Retrieved from Supply Chain Factors Dataset v1.1.,https://cfpub.epa.gov/si/si_public_record_Report.cfm?dirEntryId=349324&Lab=CESER
```
maryarbab, line 9619: Is "type" missing from the ID? same with the next one.

maryarbab, line 9619: Is "type" missing from the ID? same with the next one.

maryarbab (changes requested):

OpenEmissionFactorsDB.csv
gpantelidou: So the proposed activity_id for the 3rd one is not cement_and_concrete-concrete_[....] but similar with the other two cement_and_concrete-type_[....] ?

maryarbab: hum, my suggestion is to have building_materials-type for all three. @gpantelidou @AnastasiaLobanova thoughts?

gpantelidou: I definitely agree that it should be the same. I was also thinking that maybe in the proposed activity_id the "type" was missing.

maryarbab: Just to clarify, here are my suggestions as to be consistent with the existing cement/concrete EFs in this category: 
9617/9618- building_materials-type_cement
9619-  building_materials-type_concrete_pipe_bricks_and_blocks

maryarbab (changes requested):

OpenEmissionFactorsDB.csv
AnastasiaLobanova: Agree!

OpenEmissionFactorsDB.csv
```diff
@@ -9696,8 +9700,12 @@ b8508efa-2857-41d9-aaa3-6228be022a2f,Materials and Manufacturing,Metals,secondar
...
+f8487062-ef5b-4f10-842b-f98fb3765cc1,Materials and Manufacturing,Mined Materials,mined_materials-type_groundtreated_minerals_earth,Ground or treated minerals and earth,USD,not-supplied,not-supplied,0.552,0.003,0.0,not-supplied,0.005,,3,cradle_to_shelf,EPA,Supply Chain Factors Dataset (commodities),2022,2018,2012,US,,Climatiq,2022/10/25,Emission intensity of supply chain (with margins i.e. cradle to shelf) in US dollars spend on: ground or treated minerals and earth. This factor is representative of the described commodity (equivalent to a goods or services category) and was obtained from the USEEIO model v1.0.1. Refer to the source for the source-specific data quality assessment. Retrieved from Supply Chain Factors Dataset v1.1.,https://cfpub.epa.gov/si/si_public_record_Report.cfm?dirEntryId=349324&Lab=CESER
```
AnastasiaLobanova, line 9707: I think between ground and treated an underscore is missing

OpenEmissionFactorsDB.csv
```diff
@@ -10841,42 +10841,42 @@ d46f5fe1-8a3c-4bcf-8a8f-745504006d23,Agriculture/Hunting/Forestry/Fishing,Arable
...
+47aaa83e-1eb7-4d60-bfb8-bf5181c77afc,Agriculture/Hunting/Forestry/Fishing,Arable Farming,arable_farming_type-paddy_rice,Paddy Rice,EUR,4.2719,3.9276,not-supplied,not-supplied,not-supplied,not-supplied,not-supplied,,3,unknown,EXIOBASE,EXIOBASE 3,2021,2019,2019,AU,,Climatiq,2022/03/17,Emission intensity of supply chain in EUR spend on: paddy rice. Retrieved from the EXIOBASE v3.8.2 model outputs for products. These factors were calculated based on 2019 data. CO2 equivalent factors incorporate emissions from land use; constituent gases have not been included as they do not. The LCA boundaries of these factors are not defined by the source. These factors include effects of international trade.,https://zenodo.org/record/5589597#.Yh9_Zi8w1ao
```
AnastasiaLobanova, line 10844: The id is wrong here. The hyphen should be between farming and type:
arable_farming_type`-`paddy_rice   should read arable_farming-type_paddy_rice

AnastasiaLobanova (changes requested):

OpenEmissionFactorsDB.csv
gpantelidou: Ah yes you are right, thanks for spotting that!

OpenEmissionFactorsDB.csv
gpantelidou: yes that's true

AnastasiaLobanova: Also, for one below - I just noticed

maryarbab (approved):

AnastasiaLobanova (approved):

PR #428, opened 2023-03-28
fix activity units and update sector-category-id file

gpantelidou:

- [ ] New
- [x] Update

activity units 

**What?**

fixed activity units for 7 emission factors from GHG protocol and since it was a small PR I also updated the sector-category-id file

**Why?**

correct breaking changes and keep the sector-category-id file up-to-date


AnastasiaLobanova (commented):
Looks great! 
Did you cross-check all activity_ids structure in the structure file?

AnastasiaLobanova:
Yes, I think we should try to check in Python, I have no precise idea how but we can check tomorrow together if you want

gpantelidou:
@AnastasiaLobanova yes let's check this together tomorrow thank you :)

Sector-Category-ID_structure.csv
```diff
@@ -28,7 +29,7 @@ Consumer Goods and Services,Vehicle Maintenance and Services,consumer_goods-type
 Education,Education,education-type_[....],
 Energy,Electricity,managed_assets_electricity-energy_source_[....],
 Energy,Electricity,electricity-energy_source_[....],
-Energy,Electricity,electricity-energy_source_[....]-supplier_[�.],
+Energy,Electricity,electricity-energy_source_[....]-supplier_[....],
```
maryarbab, line 32: @gpantelidou did we already change electricity id structure?

maryarbab (changes requested):

Sector-Category-ID_structure.csv
gpantelidou: No we didn't. This id structure `electricity-energy_source_[....]-supplier_[....]` was added recently when we added electricity EFs for specific UK suppliers. In this PR just noticed this strange character �. inside the [ ]  instead of ... when I opened the file and just changed it

maryarbab: okay, let's fix this for now and then have a discussion about electricity ids

maryarbab (approved):

PR #430, opened 2023-03-31
fixed india activity_ids on AZURE 2021,2022 data to be consistent

gpantelidou:

- [ ] New
- [x] Update

Cloud Computing

**What?**

changed activity_ids and names for india AZURE 2021 data to be consistent with 2022 data

**Why?**

correct breaking chnages - be consistent between different years


gpantelidou:
thanks Maryam :)

PR #431, opened 2023-04-03
fix activity_ids for electricity residual mix EFs and update the sector-category-id file

gpantelidou:

- [ ] New
- [x] Update

Electricity

**What?**

In this PR we only change activity_ids for electricity residual mix EFs and update the sector-category-id file according to the new suggested structure.

**Why?**

We need to correct the electricity activity_ids to be consistent with each other.


OpenEmissionFactorsDB.csv
```diff
@@ -19222,71 +19222,71 @@ a4941e91-8474-4f8f-abd6-cc2f094c64a3,Energy,Electricity,electricity-energy_sourc
...
+0af58485-809e-4f6a-9cf9-ebf0efaa058a,Energy,Electricity,electricity-supply_grid-source_residual_mix_market_based,Electricity supplied from grid - residual mix,kWh,0.18906,0.18906,0.18906,not-supplied,not-supplied,not-supplied,not-supplied,,2,electricity_generation,GHG Protocol,GHG Emissions Calculation Tool,2021,2021,2019,BE,p|m,Climatiq,2022/01/15,Emission intensity of Belgium grid mix as reported. Published by the Association of Issuing Bodies (AIB). Retrieved from European Residual Mixes 2018 v1.2 (published July 2019) - Table 2 - Direct GWP (gCO2/kWh). NOTE: this value is provided in CO2 only and does not include the impact of other GHGs. NOTE: This dataset has been deprecated by the source and will be removed soon.,https://ghgprotocol.org/ghg-emissions-calculation-tool
```
maryarbab, line 19226: let's remove `market_based` from this and the following GHG protocol EFs. The `residual_mix` itself implies that this EF should be used for `market_based `reporting.

maryarbab: This change will also make the EF consistent with AIB residual mix EFs (the original source here).

maryarbab (changes requested):
Great work! Just made some small comments.

maryarbab (approved):

PR #432, opened 2023-04-06
change activity_ids for electricity production mix EFs and correct LC…

gpantelidou:
…A activities

- [ ] New
- [x] Update

Electricity, Water Treatment, Road Freight

**What?**

1. change activity_ids for electricity production mix EFs according to the new suggested structure
2. correct 10 LCA activities
3. add a `p` flag in the EF from EPPO (it already had a description note but the quality flag was missing)

**Why?**

cleanup the OEFDB


maryarbab (approved):
Looks neat!!

PR #433, opened 2023-04-06
correct activity_ids for electricity grid mix

gpantelidou:

- [ ] New
- [x] Update

Electricity

**What?**

change activity_ids for electricity grid mix & total supplier mix and small addition of p|e flag to AIB data for AT residual mix

**Why?**

correct breaking changes


maryarbab (approved):
Looks great!

PR #439, opened 2023-04-14
correct multiple errors

gpantelidou:

- [ ] New
- [x] Update

Multiple

**What?**

1. fix the rest electricity activity_ids
2. correct lca activity, scope, name, description for HKEI and CLP
3. correct ADEME source link
4. remove BEIS p and m flags
5. update the sector-category-id file
6. explain the new s flag in the data guidance file
7. fix real estate activity_ids
8. add scope 1|2 on schema file

**Why?**

necessary corrections in the OEFDB


gpantelidou:
@maryarbab,  for CLP since we took all the data from the updated file released on 2022 we end up with duplicates! What should we do now?

maryarbab:
@gpantelidou my initial comment was to include the "s" flag in the data guide doc [here](https://github.com/climatiq/Open-Emission-Factors-DB/blob/main/DATA_GUIDANCE.md). I have added a triage issue for this. But it is actually good to add it into the schema as well.

OpenEmissionFactorsDB.csv
```diff
@@ -7941,9 +7941,9 @@ ba952d4a-ee39-4e90-8975-47170703b0c6,Materials and Manufacturing,Metals,metals-t
...
+9d09af3f-4d2a-4446-b895-246cc398ad22,Energy,Electricity,electricity-supply_grid-source_supplier_mix,Electricity supplied from grid plus T&D losses,kWh,not-supplied,0.65,not-supplied,not-supplied,not-supplied,not-supplied,not-supplied,not-supplied,not-supplied,,1|2,electricity_generation-transmission_and_distribution,CLP Group,CLP Sustainability Report,2021,2021,2021,HK,,Climatiq,2023/02/06,Emission intensity of grid mix as reported (areas covered by the Hong Kong CLP Group). Retrieved from CLP Power Hong Kong Sustainability Report 2021. The emission factor is derived from the table Climate Vision 2050: CLP Group - GHG emissions intensity of generation and energy storage portfolio - On an equity basis (kg CO2e/kWh). It is combined generation and T&D factor. NOTE: The year released reported for this factor reflects the year from which the data was calculated.,https://www.clpgroup.com/en/sustainability/report-esg-ratings/sustainability-reports.html#
```
maryarbab, line 7944: I'd remove year_calculated_from unless you have actually checked all the assumptions (and their associated years) that the source has used to derive the emission factor (the concept behind year_calculated_from). This applies to both HKEI and CLP group.

maryarbab (changes requested):

OpenEmissionFactorsDB.csv
gpantelidou: but then the description note we just added should also change?

maryarbab (approved):

PR #440, opened 2023-04-17
correct BEIS 2021 activity_ids for air travel & air freight

gpantelidou:

- [ ] New
- [x] Update

Air travel and air freight

**What?**

add the sub-id `-distance_uplift_incuded` in all BEIS 2021 air travel and air freight activity_ids

**Why?**

because it is mentioned in the source methodology and to be consistent with BEIS 2022 similar data


maryarbab (commented):
Looks great!

maryarbab (approved):

PR #441, opened 2023-04-17
LCA activties

maryarbab:

- [X] New
- [X] Update

_All_

**What?**

_Add new lca_activity column and relabel the current one to source_lca_activity_

**Why?**

_Make lca_activities consistent across the OEFDB_


TimCroker:
ADEME electricity: this has been changed from "electricity generation" to "use-electricity_t&d."
I think this is correct but in which case the "Scope" column should be changed to 2 and 3.

TimCroker:
https://www.climatiq.io/data?source=ADEME&unit_type=Energy&search=f87cd572-c065-4797-8f50-4bb1098e6fc7

This has been changed from

TimCroker:
EEA for example:
4feb1df1-6953-47de-924a-9ac014dfc3c9
- these are marked as use-electricity_t&d
- I think they are just generation i.e. use only (total generation emissions / total generation) https://www.eea.europa.eu/data-and-maps/daviz/co2-emission-intensity-12#tab-chart_2

TimCroker:
"Organizational activities" from BEIS e.g. 96fa2c78-60f9-49fc-bf2f-45782ffd44cb
I think this should be the same as all spend-based factors i.e.

TimCroker:
BEIS transport factors e.g. freight, vehicles - per km / mile: these could be "use".

TimCroker (approved):

